PR #1, opened 2015-03-09
Added hasContent/Value

0x1mason:
@sebv @jlipps Please take a look at this.  This is for the xcode version changes.  I moved the the `notNullOrUndefined` functions here and renamed them to `hasContent` and `hasValue`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 module.exports = {
   tempDir: require('./lib/tempdir'),
-  system: require('./lib/system')
+  system: require('./lib/system'),
+  util: require('./lib/util')
 };
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,0 +1,22 @@
+"use strict";
+
+var _ = require('underscore');
+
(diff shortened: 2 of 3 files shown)


sebv:
`hasValue` I understand, `hasContent` I am not sure is really needed, maybe a string specific method would be better (using 1typeof1). Also could add support `bool`, cause this is the most annoying case.

0x1mason:
@sebv `bool` is supported.  Checkout the unit tests. `hasValue(false)` should be `true`.

A good example for `hasContent` is `hasContent(path) ? doSomething(path) : throw "Invalid path";

0x1mason:
Nm, I think I see your pt regaring `hasContent`, you mean keep it but take out `isNumber`.  Sure that works.

0x1mason:
@sebv Ok checkout `hasContent` now.

0x1mason:
@jlipps has seen these functions before and sebv gave his assent to the hasContent changes, so merging.  Lmk if there are concerns and I will revert.

PR #3, opened 2015-03-26
add underscore and update version

Jonahss:
will need to update package.json file in appium and other libs

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "keywords": [
     "appium"
   ],
-  "version": "0.1.1",
+  "version": "1.0.0",
   "author": "appium",
   "license": "Apache License 2.0",
   "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,7 +23,8 @@
     "lib": "lib"
   },
   "dependencies": {
-    "q": "^1.1.2"
+    "q": "^1.1.2",
+    "underscore": "^1.8.2"
```
jlipps: can we use lodash instead? i've been trying to move everything over to it

jlipps:
also, this only needs to be one commit, really.

Jonahss:
lodash now.
squashed to one commit.

jlipps:
ok, will you take care of publishing and updating downstream package.jsons?

Jonahss:
yup.

PR #4, opened 2015-03-27
add core.exec() a promise version of node child_process.exec

Jonahss:


jlipps:
can we get a test for the failure case as well? i think trying to execute a non-existing file will trigger it

jlipps:
also, i usually don't bump versions in pull requests. just when it comes time to release a new version, i run my release script and it automatically moves it to next major/minor/patch, creates a tag, publishes to npm, etc:

```
function version_and_publish() {
    name=$(cat package.json | grep '"name"' | sed -En 's/^.*"name":.+"([^"]+)".*$/\1/p')
    ver=$1
    remote=$2
    npm cache clear $name
    npm version $ver
    git push $remote master
    git push --tags $remote master
    npm publish
}
```

used like:

```
version_and_publish patch origin
```

sebv:
Q already does the wrapping for you, spread is the method you need.
```
var exec = Q.denodeify(require('child_process').exec)
exec('ls').spread(function(stdout, stderr) {
  console.log('stdout -> ' + stdout);
  console.err(stderr -> ' + stderr);
}).done()
```

jlipps:
@sebv: @Jonahss doesn't want to use Q and instead just wants to rely on native promises or a shim, that's why he's using the denodeify lib instead of Q

sebv:
I don't really see why, since Q is already in package.json, but in any case `spread` is the correct pattern, it takes 3 or 4 lines to reimplement it.

sebv:
I mean you don't want to write a wrapper for every callback with more than one no-error param.

jlipps:
seems like `spread` makes it hard to get the results as part of an `await` assignment. we want to be able to do:

```
let {stdout, stderr} = await exec('ls');
```

sebv:
You can do:
```
let [stdout, stderr] = await exec('ls');
```

jlipps:
what is the wrapped function for that? is it this?

```
var exec = Q.denodeify(require('child_process').exec)
exec('ls').spread(function(stdout, stderr) {
  return [stdout, stderr];
});
```

just wondering how spread works, seems like it already resolves the promise...

sebv:
Just `denodeify` you can try this, res is an array.
```
var Q = require('q'),
    exec = Q.denodeify(require('child_process').exec);
exec("ls").then(function(res) {
  console.log('res ->', res);
}).done();
```

sebv:
more or less this is spread:

```js
p.spread = function(f) {
  this.then(function(res) {
    if(!(res instanceof Array)) res = [res];
    f.apply(null, res);
  });
}
```

sebv:
babel example:

```
require("babel/polyfill");

import Q from 'q';
import cp from 'child_process';

let exec = Q.denodeify(cp.exec);

let f = async function() {
  let [stdout, stderr] = await exec('ls');
  console.log('stdout ->', stdout);
  console.log('stderr ->', stderr);
}

Q(f()).done();
```

Jonahss:
We don't need to create a wrapper like this every time we use a function like this.
I just thought this one would be simpler than wrapping exec all the time, since we use it everywhere.

They way it's done with denodeify it only comes out as 1-2 lines anyways.

Jonahss:
Added a test for failing

sebv:
Well for the record I am :-1: on this, the reason are:

- We don't need to include extra promise libraries
- For es7 we don't need it
- For es5 we should use Q.spread, or use a pattern which is functional (i.e. as in functional programming).

If we want to switch promise library we should choose a better one like bluebird, otherwise we will end up having to re-implement  all the helpers.

jlipps:
Jonah, what do you think about sebv's points?

Jonahss:
Since es6 and es7 have native promises, I want to use native promises as much as I can instead of using promise libraries like `q`. Denodeify isn't a promise library, just a utility, but in order to work on es5 it needs a promise library. Right now it loads `native-promise-only` to satisfy that. I really don't want to include `q` in a library that's written for es7.

I can see the value of only using one promise library though, for consistency.

Jonahss:
I'm fine with just rewriting a promise version of the exec function in every file I use it. Then can use Q in es5 and denodeify in es7

sebv:
The best would actually be to drop Q and not using the the native promise, neither are competitive, one is too old, one is too young. The good ones are bluebird, and the RSVP. But your idea of just using native promise right now is a bit crazy, there is not enough functionality in the A+ promises. You should wait for the ecosystem to settle, very likely a standard will emerge enabling those libs to extend them in a cheap way.

sebv:
@Jonahss you should have very little use for this function, it's not good to transform arrays into object if you don't need to, javascript is much better at manipulating array. I had the same issue before with function with several return values and promises, but my brain adapted to it.

Jonahss:
I really wanted this function *just* for exec. I'm ok with doing it in-line.

If we are using es7 with Traceur I think we should use the native promises 100%. If you think we should wait for the ecosystem to settle , and that the features are too young, we shouldn't be using Traceur at all.

sebv:
I don't really see a tight link between promise and traceur, you can actually make an objective choice to which is the best library to use, and native promise is not the best choice, neither is Q.

Actually it is probably a good strategy to limit exposure to es7 in term of external library because this code is not transpiled.

Jonahss:
I was thinking along the same lines. Could we create a separate
appium-support module just for es7?

Jonahss:
changing this to just return a denodeified version of exec() but not wrapping the return values in an object. just returning them as an array.

Jonahss:
@sebv good to merge?

Jonahss:
@sebv replaced Q with bluebird.

PR #5, opened 2015-03-31
promise-based fileExists() function

Jonahss:


Jonahss:
changed the "if/else" to just `resolve(!err)`

test/util-specs.js
```diff
@@ -47,4 +47,24 @@ describe('util', function () {
       util.hasContent(e).should.be.false;
     });
   });
+
+  describe("fileExists", function () {
+    it("should return true if file is readable", function (done) {
```
sebv: drop the done, you can just return the promise.

PR #6, opened 2015-04-01
Nodeify module

Jonahss:
This code just copied from future.js in appium, so that we don't have to copy code between different future.js files in different appium modules

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -31,3 +31,21 @@ exports.fileExists = function (path) {
     });
   });
 };
+
+exports.nodeifyModule = function (obj) {
+  if (typeof obj !== "function") {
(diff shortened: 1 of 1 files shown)


jlipps:
please squash to one commit

Jonahss:
@jlipps does that mean that after squashing it can be merged?

For clarity's sake I'd like to submit PRs as multiple commits, then squash before merging. Is that an acceptable approach going forward?

jlipps:
yep, LGTM after squashing. your approach sounds fine. i'll try to remember to give all other kinds of feedback before asking to squash

Jonahss:
lol. LGTS

PR #7, opened 2015-04-06
Added parse path method

0x1mason:
Needed for some logging changes I'm making.

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -1,5 +1,7 @@
 "use strict";
 
+var _ = require('underscore')
+
 var osType = require('os').type();
 
 exports.isWindows = function () {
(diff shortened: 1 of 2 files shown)


lib/system.js
```diff
@@ -12,5 +14,86 @@ exports.isMac = function () {
...
+  /^([a-zA-Z]:|[\\\/]{2}[^\\\/]+[\\\/]+[^\\\/]+)?([\\\/])?([\s\S]*?)$/;
+
+// Regex to split the tail part of the above into [*, dir, basename, ext]
+var splitTailRe =
+  /^([\s\S]*?)((?:\.{1,2}|[^\\\/]+?|)(\.[^.\/\\]*|))(?:[\\\/]*)$/;
+
+// Function to split a filename into [root, dir, basename, ext]
+function win32SplitPath(filename) {
+  // Separate device+slash from tail
+  var result = splitDeviceRe.exec(filename),
```
jlipps: when doing multiple instantiations like this, please indent subsequent lines 4 spaces (so they line up to the right of `var`)

lib/system.js
```diff
@@ -12,5 +14,86 @@ exports.isMac = function () {
...
+    device = (result[1] || '') + (result[2] || ''),
+    tail = result[3] || '';
+  // Split the tail into dir, basename and extension
+  var result2 = splitTailRe.exec(tail),
+    dir = result2[1],
+    basename = result2[2],
+    ext = result2[3];
+  return [device, dir, basename, ext];
```
jlipps: I like how you returned an object below, instead of an array, can we do that here?

lib/system.js
```diff
@@ -12,5 +14,86 @@ exports.isMac = function () {
...
+}
+
+exports.win32ParsePath = function(pathString) {
+  if (!_.isString(pathString)) {
+    throw new TypeError(
```
jlipps: our style for breaking long strings would do something like:

```js
throw new TypeError("Parameter 'pathString' must be a strong, not " +
                    typeof pathString);
```

lib/system.js
```diff
@@ -12,5 +14,86 @@ exports.isMac = function () {
...
+    root: allParts[0],
+    dir: allParts[0] + allParts[1].slice(0, -1),
+    base: allParts[2],
+    ext: allParts[3],
+    name: allParts[2].slice(0, allParts[2].length - allParts[3].length)
+  };
+};
+
+
+if (exports.isWindows())
```
jlipps: please use braces even with one-line blocks. The only exception for this is in the case where we're short-circuiting a callback chain, as in:

```js
if (err) return cb(err);
```

test/system-specs.js
```diff
@@ -11,4 +11,20 @@ describe('system', function () {
     should.exist(system.isLinux);
     should.exist(system.isMac);
   });
+
+  it('parse paths properly', function () {
```
jlipps: please change the spec to include the word `should` to start off

jlipps:
made a few comments, looks good. what about putting these in their own `path.js` instead of here in system?

lib/system.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+var _ = require('underscore')
```
jlipps: semicolon. did this pass lint?

lib/system.js
```diff
@@ -12,5 +14,86 @@ exports.isMac = function () {
...
+// tail] windows-only
+var splitDeviceRe =
+  /^([a-zA-Z]:|[\\\/]{2}[^\\\/]+[\\\/]+[^\\\/]+)?([\\\/])?([\s\S]*?)$/;
+
+// Regex to split the tail part of the above into [*, dir, basename, ext]
+var splitTailRe =
+  /^([\s\S]*?)((?:\.{1,2}|[^\\\/]+?|)(\.[^.\/\\]*|))(?:[\\\/]*)$/;
+
+// Function to split a filename into [root, dir, basename, ext]
+function win32SplitPath(filename) {
```
jlipps: `function win32SplitPath (filename) {`

probably need to run lint, might be other things

PR #12, opened 2015-07-07
Sonia rewrite appium-support

syang97:


index.js
```diff
@@ -1,6 +1,10 @@
...
+
+import tempDir from './lib/tempdir';
+import system from './lib/system';
+import util from './lib/util';
+import core from './lib/core';
+
+export default tempDir; 
+export default system;
+export default util;
+export default core;
```
imurchie: Is this possible? I thought there could only be a single default. And what does it even mean, to have four defaults?

lib/core.js
```diff
@@ -1,6 +1,7 @@
-"use strict";
 
-var child_process = require('child_process')
-  , B = require('bluebird');
+import child_process from 'child_process';
```
imurchie: We should be using [teen_process](https://github.com/appium/node-teen_process). That way you won't need to `promisify` or anything. It doesn't look like we will need this `core` module at all, in fact.

lib/system.js
```diff
@@ -1,27 +1,33 @@
-"use strict";
 
-var exec = require('child_process').exec
-  , osType = require('os').type();
+import { exec } from 'teen_process';
+import os from 'os';
+import { getLogger } from 'appium-logger';
+let osType = os.type();
```
imurchie: Can you organize these so the `os` imports are together?

lib/tempdir.js
```diff
@@ -1,17 +1,19 @@
...
-var fs   = require('fs')
-  , os   = require('os')
-  , path = require('path')
-  , cnst = require('constants')
-  , B = require('bluebird');
+const logger = getLogger('Appium-Support');
+let promisifiedFsOpen = B.promisify(fs.open);
 
-var RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
+let RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
```
imurchie: Since this is a constant, maybe use `const`?

lib/system.js
```diff
@@ -1,27 +1,33 @@
...
-exports.isWindows = function () {
+const logger = getLogger('Appium-Support');
+
+export function isWindows () {
   return osType === 'Windows_NT';
-};
+}
 
-exports.isMac = function () {
+export function isMac () {
```
imurchie: I think we, as a team, need to standardize how we are doing exports. Should they be on individual functions (as here), or using an `export { ... };` call, as in `tempdir.js` below. In my opinion it should be the latter, so that at a glance the exports from a module are visible.

lib/tempdir.js
```diff
@@ -1,17 +1,19 @@
 /* This library is originated from temp.js at http://github.com/bruce/node-temp */
-"use strict";
+import fs from 'fs';
+import os from 'os';
+import nodePath from 'path';
```
imurchie: I think `path` would be a better name for this. It's what is used elsewhere, and makes it so we know what it is when reading the code, without needing to look up at the imports.

Oh! I see we export a `path` function. I would still prefer to use `path` here, for consistency's sake, and then rename our `path` and just export it as `path`. That is

```javascript
...
function getPath (rawAffixes, defaultPrefix) {
  let affixes = parseAffixes(rawAffixes, defaultPrefix);
  let name = [affixes.prefix, affixes.suffix].join('');
  return nodePath.join(tempDir(), name);
}
...
exports (open, getPath as path);
```

I don't entirely know if that works, but it ought to ;)

index.js
```diff
@@ -1,6 +1,10 @@
...
-  util: require('./lib/util'),
-  core: require('./lib/core')
-};
+
+import tempDir from './lib/tempdir';
+import system from './lib/system';
+import util from './lib/util';
+import core from './lib/core';
+
+export * as tempDir;
```
sebv: This doesn't right, the same thing is exported 4 times with different names.

lib/tempdir.js
```diff
@@ -21,25 +23,29 @@ var tempDir = function () {
...
 
 
-var open = function (affixes, callback) {
-  var filePath = generateName(affixes, 'f-');
-  fs.open(filePath, RDWR_EXCL, 384, function (err, fd) {
-    if (callback)
-      callback(err, {path: filePath, fd: fd});
-  });
-};
+let open = async function (affixes) {
```
imurchie: `async function open (affixes)`

lib/tempdir.js
```diff
@@ -21,25 +23,29 @@ var tempDir = function () {
...
+  let filePath = path(affixes, 'f-'); 
+  try {
+    let fd = await promisifiedFsOpen(filePath, RDWR_EXCL, 384);
```
imurchie: Can we get a comment on `384`? That seems magical.

lib/system.js
sebv: Just move the method call inside the functions where it is used..

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
+import { getLogger } from 'appium-logger';
 
+let logger = getLogger('Appium-Support');
+
+let promisifiedExists = B.promisify(fs.exists);
+let promisifiedMkdir = B.promisify(fs.mkdir);
+let promisifiedAccess = B.promisify(fs.access);
 // return true if the the value is not undefined, null, or empty string.
-exports.hasContent = function (val) {
+export function hasContent (val) {
```
imurchie: Ditto `export` comment above.

lib/system.js
```diff
@@ -1,27 +1,33 @@
...
-exports.arch = function (cb) {
-  exec("uname -m", function (err, stdout) {
-    if (err) return cb(err);
-    if (stdout.trim() === "i686") {
-      cb(null, "32");
+export async function arch () { 
+  try {
+    let {stdout} = await exec('uname', ['-m']);
+    
+    if (stdout.trim() === 'i686') {
```
sebv: return stdout.trim() === 'i686' ? '32' : '64';

lib/tempdir.js
```diff
@@ -1,17 +1,19 @@
...
+import B from 'bluebird';
+import { getLogger } from 'appium-logger';
 
-var fs   = require('fs')
-  , os   = require('os')
-  , path = require('path')
-  , cnst = require('constants')
-  , B = require('bluebird');
+const logger = getLogger('Appium-Support');
+let promisifiedFsOpen = B.promisify(fs.open);
```
sebv: I think like this it is better:
```
import _fs from 'fs';
let fs = {
    open:  B.promisify(_fs.open); 
}
```

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
-  , B = require('bluebird');
+import _ from 'lodash';
+import fs from 'fs';
+import os from 'os';
+import B from 'bluebird';
+import { getLogger } from 'appium-logger';
 
+let logger = getLogger('Appium-Support');
+
+let promisifiedExists = B.promisify(fs.exists);
```
sebv: This doesn't work, they forgot the error in the exists api, you need to wrap it yourself.

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
+let promisifiedMkdir = B.promisify(fs.mkdir);
+let promisifiedAccess = B.promisify(fs.access);
 // return true if the the value is not undefined, null, or empty string.
-exports.hasContent = function (val) {
+export function hasContent (val) {
   return _.isString(val) && val !== "";
```
sebv: Just wondering why we ever need this method?

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
+    await promisifiedMkdir(dirName);
+  } catch (err) {
+    if (err && err.code !== "EEXIST") {
+      logger.errorAndThrow(err);
+    } 
+  }
+}
 
-exports.nodeifyModule = function (obj) {
+export function nodeifyModule (obj) {
```
sebv: If we use it only in 1.5 we don't need this logic.

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
-        resolve(exists);
-      });
-    }
 
-    fs.access(path, fs.F_OK | fs.R_OK, function (err) {
-      resolve(!err);
-    });
-  });
-};
+  if (!fs.access) {
```
imurchie: Since we're requiring node 0.12 now, I don't think this part if necessary. 

If it is necessary, what are the consequences of running `B.promisify` on `undefined`, which would happen above?

package.json
```diff
@@ -17,30 +17,32 @@
...
-    "rimraf": "^2.4.0",
-    "run-sequence": "^1.0.2"
+    "appium-gulp-plugins": "^1.2.12",
+    "chai": "^3.0.0",
+    "chai-as-promised": "^5.1.0",
+    "gulp": "^3.8.11",
+    "mochawait": "^2.0.0",
+    "yargs": "^3.10.0",
+    "rimraf": "^1.0.0",
+    "teen_process": "^1.0.0",
```
sebv: not a dev dep.

test/system-specs.js
```diff
@@ -12,12 +11,9 @@ describe('system', function () {
...
   if (system.isLinux || system.isMac) {
-    it('should know architecture', function (done) {
-      system.arch(function (err, arch) {
-        should.not.exist(err);
-        ["32", "64"].should.contain(arch);
-        done();
-      });
+    it('should know architecture', async function () {
+      let arch = await system.arch();
+      ["32", "64"].should.contain(arch);
```
sebv: Just make it return what you need.

test/system-specs.js
```diff
@@ -12,12 +11,9 @@ describe('system', function () {
     should.exist(system.isMac);
```
sebv: We wrote a bad test because we didn't have the habit of writing unit test at the time, but this should be improved.

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
+  } catch (err) {
+    if (err && err.code !== "EEXIST") {
+      logger.errorAndThrow(err);
+    } 
+  }
+}
 
-exports.nodeifyModule = function (obj) {
+export function nodeifyModule (obj) {
   if (typeof obj !== "function") {
```
imurchie: Elsewhere you use `lodash` methods for determining types. I think that should continue here. That is, `if (!_.isFunction(obj)) {`

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "appium-support",
-  "description": "Support libs used across appium packages.",
+  "description": "Description goes here.",
```
imurchie: Can we get a description?

package.json
```diff
@@ -1,10 +1,10 @@
...
   "keywords": [
     "appium"
   ],
-  "version": "1.1.1",
+  "version": "0.0.0",
```
imurchie: `2.0.0`

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
+      
+       
     });
   });
 
   describe("localIp", function () {
     it("should find a local ip address", function () {
-      var ip = util.localIp();
+      let ip = util.localIp();
       ip.should.match(/\d*\.\d*\.\d*\.\d*/);
```
sebv: run the logic once, see what it return and stub/mock it, that'll be more solid.

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
-            exists.should.be.true;
-            done();
-          }).catch(done);
-        }).catch(done);
-      });
+    let dirName = path.resolve(__dirname, "tmp");
+    
+    it("should make a directory that doesn't exist", async function () {
+      try { 
+        promisifiedRimraf(dirName);
```
sebv: await missing.
`rimraf` would be better than `promisifiedRimraf`

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
+      } catch (err) { 
+        logger.errorAndThrow(err);
+      }
     });
 
-    it("should still throw an error if something else goes wrong", function (done) {
-      util.mkdirp("/bin/foo").should.eventually.be.rejectedWith('EACCES')
-                             .then(done).catch(done);
+    it("should still throw an error if something else goes wrong", function () {
+      util.mkdirp("/bin/foo").should.eventually.be.rejectedWith('EACCES');
```
sebv: await

test/tempdir-specs.js
```diff
@@ -1,27 +1,25 @@
...
+import fs from 'fs';
+import chai from 'chai';
 
 chai.should();
 
 describe('tempdir', function () {
   it('should be able to generate a path', function () {
-    var path = tempDir.path({prefix: 'myfile', suffix: '.tmp'});
+    let path = tempDir.path({prefix: 'myfile', suffix: '.tmp'});
     path.should.exists;
```
imurchie: For grammar's sake, `path.should.exist`.

lib/util.js
sebv: How about the sauce publish script? It is using 0.10, does it need to be updated?

test/util-specs.js
```diff
@@ -1,16 +1,20 @@
...
+import chaiAsPromised from 'chai-as-promised';
+import chai from 'chai';
+import { getLogger } from 'appium-logger';
+import B from 'bluebird';
+
+let promisifiedRimraf= B.promisify(rimraf);
+const logger = getLogger('Appium-Support');
 chai.use(chaiAsPromised);
-var should = chai.should();
+let should = chai.should();
```
imurchie: We can just do `chai.should();` and change the instances of `should(x)...` to `x.should...`.

test/util-specs.js
```diff
@@ -1,16 +1,20 @@
...
 
 describe('util', function () {
-  var nan = NaN
+  let nan = NaN
```
imurchie: I don't quite get the utility of these being declared in this way. It makes the tests more difficult to reason about at a glance, since at the time of their use it's not obvious that, for instance, `n` means `0`.

imurchie: Since it's not part of this diff, I'll add it here:

I would also prefer the tests for `hasValue` and `hasContent` to be split up so that each value is in its own test. This will mean more tests, but it also means that we test them all on failure of one. Rather than `"should work as expected"`, I would prefer `"should handle NaN"`, `"should handle 0"`, etc etc.

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
-            done();
-          }).catch(done);
-        }).catch(done);
-      });
+    let dirName = path.resolve(__dirname, "tmp");
+    
+    it("should make a directory that doesn't exist", async function () {
+      try { 
+        promisifiedRimraf(dirName); 
+      } catch (err) {
```
imurchie: Do we need to catch the error? If it throws an error, it is going to fail the test anyway, and the test runner will log the output.

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
+        logger.errorAndThrow(err); 
+      }
+      await util.mkdirp(dirName);
+      try {
+        let exists = await util.fileExists(dirName);
+        exists.should.be.true;
+      } catch (err) {
```
imurchie: ditto.

test/util-specs.js
```diff
@@ -53,73 +57,77 @@ describe('util', function () {
...
-      util.fileExists(dirName).then(function (exists) {
+    
+    it("should not complain if the dir already exists", async function () {
+      try {
+        let exists = await util.fileExists(dirName);
         exists.should.be.true;
-        util.mkdirp(dirName).then(done).catch(done);
-      }).catch(done);
+        await util.mkdirp(dirName);
+      } catch (err) {
```
imurchie: ditto

imurchie:
This is shaping up!

test/util-specs.js
sebv: Yeah it looks like left out debugging, totally overkill in specs.

lib/util.js
```diff
@@ -1,68 +1,74 @@
...
   return str.split(/ /).join('\\ ');
-};
+}
 
 // promise which resolves true if program has access to read the path passed in,
 // false otherwise
-exports.fileExists = function (path) {
-  return new B(function (resolve) {
-
+export async function fileExists (path) {
```
imurchie: This function name doesn't seem to reflect what the comment on the function says it does. It would be better to name it something about having read access.

lib/tempdir.js
```diff
@@ -1,17 +1,23 @@
...
 
-var fs   = require('fs')
-  , os   = require('os')
-  , path = require('path')
-  , cnst = require('constants')
-  , B = require('bluebird');
+const logger = getLogger('Appium-Support');
+let fs = {
+  open: B.promisify(_fs.open),
+  mkdirSync: _fs.mkdirSync,
```
sebv: Avoid using the sync methods, they are blocking.

lib/tempdir.js
```diff
@@ -1,17 +1,23 @@
...
+  existsSync: _fs.existsSync
```
sebv: Same, for exists you can do something like:
exist: promisify(function(path, done) {_fs.exists (path, done.bind(null, null); }),

sebv:
@syang97 missing dependencies here as well, check the Travis build below.

lib/system.js
```diff
@@ -1,27 +1,40 @@
...
+  if (isLinux() || isMac()) {
+    try {
+      let {stdout} = await exec('uname', ['-m']);
+      return stdout.trim() === 'i686' ? '32' : '64';
+    } catch (err) {
+      logger.errorAndThrow(err); 
     }
-  });
-};
+  }
```
jlipps: we do `} else if ... {` (on the same line)

lib/system.js
```diff
@@ -1,27 +1,40 @@
...
-      cb(null, "32");
-    } else {
-      cb(null, "64");
+async function arch () { 
+  if (isLinux() || isMac()) {
+    try {
+      let {stdout} = await exec('uname', ['-m']);
+      return stdout.trim() === 'i686' ? '32' : '64';
+    } catch (err) {
+      logger.errorAndThrow(err);
```
jlipps: this is just my opinion but I feel that appium-support helper functions are too low-level to need logging. what do you think @sebv @imurchie ?

lib/system.js
```diff
@@ -1,27 +1,40 @@
...
+      return stdout.trim() === 'i686' ? '32' : '64';
+    } catch (err) {
+      logger.errorAndThrow(err); 
     }
-  });
-};
+  }
+ else if (isWindows()) {
+  try { 
+    let {stdout} = await exec('echo PROCESSOR_ARCHITECTURE');
```
jlipps: ah, so this is what you do on windows (cc @moizjv).

also, spacing seems funky here

lib/tempdir.js
jlipps: yep, this is the nice thing about `async/await`--we can use promisified methods as though they are sync methods, even though they are still async.

sync methods actually block javascript's event loop so we are removing all usage of them throughout appium

lib/tempdir.js
```diff
@@ -21,25 +27,31 @@ var tempDir = function () {
...
+let open = async function (affixes) {
+  let filePath =  path(affixes, 'f-'); 
+  try {
+    let fd = await fs.open(filePath, RDWR_EXCL, 384);
 
-var parseAffixes = function (rawAffixes, defaultPrefix) {
-  var affixes = {prefix: null, suffix: null};
+    return {path: filePath, fd: fd};
+  } catch (err) { 
+    console.log('No');
```
jlipps: i'm guessing this is for debugging? :-)

lib/tempdir.js
```diff
@@ -55,7 +67,6 @@ var parseAffixes = function (rawAffixes, defaultPrefix) {
     affixes.prefix = defaultPrefix;
   }
   return affixes;
-};
+}
 
-exports.open = B.promisify(open);
-exports.path = generateName;
+export { open, path };
```
jlipps: yep, I agree with isaac. We should have all exports at the end of the file.

Also, I think `export as ...` is broken in unsupported in babel currently.

lib/util.js
```diff
@@ -1,85 +1,68 @@
...
-      });
-    }
-
-    fs.access(path, fs.F_OK | fs.R_OK, function (err) {
-      resolve(!err);
-    });
-  });
-};
+  try { 
+    console.log('path' + path);
```
jlipps: remove debug statement

lib/util.js
```diff
@@ -88,4 +71,6 @@ exports.localIp = function () {
     .pluck('address')
     .first().value();
```
jlipps: every other call is on its own line so let's fix this up to be the same:

```js
.first()
.value();
```

lib/system.js
sebv: Throwing exception is good enough, I don't like systematic `errorAndThrow`.

sebv:
Could we also add this to export (and test it):
 ```sj
const fs = {
  open: B.promisify(_fs.open),
  close: B.promisify(_fs.close),
   ...
};
```
Feel like we are duplicating this everywhere

File: lib/system.js (hunk 36 above)
Comment (imurchie) [inline lib/system.js]: I agree. The calling code can handle logging in a context that makes more sense.

File: lib/util.js [hunk 42]
```diff
@@ -1,85 +1,68 @@
...
-exports.nodeifyModule = function (obj) {
-  if (typeof obj !== "function") {
-    var nodeified = {};
-    _.each(Object.getOwnPropertyNames(obj), function (name) {
-      if (name !== "__esModule") {
-        nodeified[name] = exports.nodeifyModule(obj[name]);
-      }
-    });
-    return nodeified;
+async function mkdirp (dirName) {
```
imurchie: This function seems to have a disconnect between the name and what it does. `mkdirp` suggests it will function like `mkdir -p` (i.e., recursive directory creation) while `fs.mkdir()` will only create a single level, and throw an `ENOENT` error if you try to create multiple levels. 

It might be that we only want to create a single level of directory, which is fine, but then we ought to just have `mkdir` as the name. Otherwise perhaps we need to use a library like [node-mkdirp](https://github.com/substack/node-mkdirp) to get the correct functionality.

lib/util.js
```diff
@@ -1,85 +1,68 @@
...
+  try { 
+    await fs.mkdir(dirName);
+  } catch (err) {
+    if (err && err.code !== "EEXIST") {
```
imurchie: What about other errors?

test/system-specs.js
```diff
@@ -1,9 +1,9 @@
-"use strict";
+'use strict';
+import * as system from '../lib/system';
+//import os from 'os';
```
imurchie: Just get rid of it if it's not used.

test/util-specs.js
```diff
@@ -1,16 +1,19 @@
...
-  , chai = require('chai');
+import * as util from '../lib/util';
+import rimraf from 'rimraf';
+import path from 'path';
+import chai from 'chai';
+import { getLogger } from 'appium-logger';
+import B from 'bluebird';
+
+let promisifiedRimraf= B.promisify(rimraf);
+const logger = getLogger('Appium-Support');
```
imurchie: All of the logging in the `specs` should be left to the test runner. Just allow any exceptions to go, and reports will be made in the expected format.

test/system-specs.js
syang97: @sebv not sure what you mean by that. Do you mean compact it in one line like `['32', '64'].should.contain(await system.arch()?`

test/util-specs.js
syang97: This doesn't work unless rimraf is promisified though

sebv: `rimraf = B.promisify(`_rimraf`)`

lib/tempdir.js
syang97: Hey sebv, that didn't seem to work. I tried something like 
`let fs = {
  open: B.promisify(_fs.open),
  mkdir: B.promisify(_fs.mkdir),
  exists:  B.promisify(function (path, done) {
                  _fs.exists (path, function (result) {
                        done(null, result);
                      });
                })
}; ` 
which also didn't work

imurchie: I think we would be better off following the advice of the authors of the function:

>fs.exists() is an anachronism and exists only for historical reasons. There should almost never be a reason to use it in your own code.

>In particular, checking if a file exists before opening it is an anti-pattern that leaves you vulnerable to race conditions: another process may remove the file between the calls to fs.exists() and fs.open(). Just open the file and handle the error when it's not there.

Only the logic would be the other way around (i.e., making the directory and handling the error if it already exists).

sebv: I think the authors are talking about opening file descriptors rather than opening files, which won't fix the race condition on unix systems.  You should do both: 
- Check for existence using file descriptor (or a proper exist method implemented with file descriptors)
- Catch the exception on file/dir creation. (if you actually need to create a file, which is not the most common use case.)

sebv: @syang97 the code seem to work as expected: https://gist.github.com/sebv/a4cb31595c46b6648f18

It'll be better to use `bind` thought.

lib/system.js
```diff
@@ -1,27 +1,36 @@
...
-var exec = require('child_process').exec
-  , osType = require('os').type();
+import { exec } from 'teen_process';
+import os from 'os';
 
-exports.isWindows = function () {
+
+function isWindows () {
+  let osType = os.type();
   return osType === 'Windows_NT';
```
imurchie: Does having `osType` give us any advantages, over just using `os.type()`? 

```javascript
return os.type() === 'Windows_NT';
```

lib/system.js
```diff
@@ -1,27 +1,36 @@
...
-};
+}
 
-exports.isMac = function () {
+function isMac () {
+  let osType = os.type();
   return osType === 'Darwin';
```
imurchie: Ditto.

lib/system.js
```diff
@@ -1,27 +1,36 @@
...
-};
+}
 
-exports.isLinux = function () {
+function isLinux () {
   return !exports.isWindows() && !exports.isMac();
```
imurchie: I don't think you want `exports` here. Just `return !isWindows() && !isMac();`

lib/system.js
```diff
@@ -1,27 +1,36 @@
...
-      cb(null, "32");
-    } else {
-      cb(null, "64");
+async function arch () { 
+  if (isLinux() || isMac()) {
+    try {
+      let {stdout} = await exec('uname', ['-m']);
+      return stdout.trim() === 'i686' ? '32' : '64';
+    } catch (err) {
+      throw err;
```
imurchie: No need for the `try`/`catch` block, since you're just catching and throwing the same `err` object, with no processing.

lib/tempdir.js
```diff
@@ -1,45 +1,63 @@
...
+import B from 'bluebird';
+import { getLogger } from 'appium-logger';
 
-var fs   = require('fs')
-  , os   = require('os')
-  , path = require('path')
-  , cnst = require('constants')
-  , B = require('bluebird');
+const logger = getLogger('Appium-Support');
+import { hasAccess } from '../lib/util';
```
imurchie: We are in `lib`, so this can just be `'./util`.

lib/tempdir.js
```diff
@@ -1,45 +1,63 @@
...
+let fs = {
+  open: B.promisify(_fs.open),
+  mkdir: B.promisify(_fs.mkdir),
+  exists: B.promisify(
```
imurchie: After all of that, this is never used!

lib/tempdir.js
```diff
@@ -1,45 +1,63 @@
...
+    })
+};
 
-var RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
+const RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
 
-var tempDir = function () {
-  var now = new Date();
-  var filePath = path.join(os.tmpDir(),
+async function tempDir () {
```
imurchie: This function could use some documentation about the naming scheme of the directory created/found.

lib/tempdir.js
```diff
@@ -55,7 +73,6 @@ var parseAffixes = function (rawAffixes, defaultPrefix) {
     affixes.prefix = defaultPrefix;
```
imurchie: On line 70 above (not part of this diff) there is

```javascript
throw ("Unknown affix declaration: " + affixes);
```

It would be better to throw an `Error`, and use interpolation. I.e., 

```
throw new Error(`Unknown affix declaration: ${affixes}`);
```

lib/util.js
```diff
@@ -1,91 +1,75 @@
...
-      if (err && err.code !== "EEXIST") {
-        reject(err);
-      } else {
-        resolve();
-      }
-    });
-  });
-};
+  // TODO update requirements to node v0.12. Until then, fs.access might
+  // not be defined
```
imurchie: This is no longer a TODO.

package.json
```diff
@@ -17,30 +17,33 @@
...
   },
   "dependencies": {
-    "bluebird": "^2.9.21",
-    "lodash": "^3.6.0"
+    "babel-runtime": "=5.5.5",
+    "source-map-support": "^0.3.1",
+    "appium-logger":"1.0.0",
+    "teen_process": "^1.0.0",
+    "bluebird": "^2.9.25",
+    "lodash": "^1.0.0",
```
imurchie: Why such an old version of `lodash`? `3.10.0` is the current version.

test/tempdir-specs.js
```diff
@@ -1,27 +1,26 @@
...
-      .then(function (res) {
-        res.should.exist;
-        res.path.should.exist;
-        res.path.should.include('my-test-file.zip');
-        res.fd.should.exist;
-        fs.existsSync(res.path).should.be.ok;
-      });
+  it('should be able to create a temp file', async function () {
+    let res = await tempDir.open({prefix: 'my-test-file', suffix: '.zip'});
+    console.log('after');
```
imurchie: I think this is not necessary ;)

lib/system.js
```diff
@@ -1,27 +1,29 @@
...
-};
+function isWindows () {
+  return os.type() === 'Windows_NT';
+}
 
-exports.isMac = function () {
-  return osType === 'Darwin';
-};
+function isMac () {
+  return os.type === 'Darwin';
```
imurchie: `os.type` needs to be called.

lib/system.js
```diff
@@ -1,27 +1,29 @@
...
+function isOSWin64() {
+  return process.arch === 'x64' || process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
+}
+
+async function arch () { 
+  if (isLinux() || isMac()) {
+      let {stdout} = await exec('uname', ['-m']);
+      return stdout.trim() === 'i686' ? '32' : '64';
+    } else if (isWindows()) {
+   return isOSWin64 ? '64' : '32';
```
imurchie: indentation.

lib/system.js
imurchie: This also points to the need for a test to make sure that these three functions work correctly. Currently there is a test to make sure that the three functions exist (not sure about the utility of that test), one to test that `arch` works, and one to test that one of the three functions above returns `true`. 

I think that the entirety of [that spec](https://github.com/syang97/appium-support/blob/sonia-rewrite/test/system-specs.js) could be improved:

1. `isWindows`, `isMac`, `isLinux` - stub `os.type` to return `Windows_NT`, `Darwin`, and something else, respectively, to guarantee the logic works
2. `arch` - test the four different cases (Mac/Linux 32 & 64 bit architectures, Windows 32 & 64 bit architectures), through stubbing/mocking `os.type` and `exec`.

test/system-specs.js
```diff
@@ -1,9 +1,17 @@
...
-var should = chai.should();
+let should = chai.should();
+
+let sandbox = null;
+let SANDBOX = Symbol();
+let mocks = {};
+let libs = {teen_process, os, system};
 
 describe('system', function () {
   it('should exist', function () {
```
imurchie: I don't think we need these tests at all. They are testing that the functions are there, which is not really useful.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
   });
-  if (system.isLinux || system.isMac) {
-    it('should know architecture', function (done) {
-      system.arch(function (err, arch) {
-        should.not.exist(err);
-        ["32", "64"].should.contain(arch);
-        done();
-      });
+
+  if (system.isLinux || system.isMac || system.isWindows) {
```
imurchie: Ditto this test, which is now subsumed within the ones below exercising `arch` more directly.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
+    it('should know architecture', async function () {
+      let arch = await system.arch();
+      ["32", "64"].should.contain(arch);
+    });
+  } 
+
+  it('should detect the operating system', function () {
```
imurchie: Ditto this test.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
+    osMock.restore();
+  });
+
+  it('should correctly return Mac if it is a Mac', function () { 
+    let osMock = sinon.mock(os);
+    osMock.expects('type').returns('Darwin');
+    system.isMac().should.equal.true;
+    osMock.verify();
+    osMock.restore();
+  });
```
imurchie: Need a test for `isLinux`, even though it just delegates to `isMac` and `isWindows`.

imurchie: I would also put all three into a `describe` block, too (something like `describe("isX functions", ...`), just for grouping's sake.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
     });
-  }
+
+    it('should return correct architecture if it is a 64 bit Windows', async function() {
+      mocks.os.expects('type').thrice().returns('Windows_NT');
+      mocks.system.expects('isOSWin64').once().returns(true);
+      let arch = await system.arch();
+      arch.should.equal('64');
+      mocks[SANDBOX].verify();
+  });
```
imurchie: Indentation.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
+
+    it('should return correct architecture if it is a 32 bit Windows', async function() {
+      mocks.os.expects('type').thrice().returns('Windows_NT');
+      mocks.system.expects('isOSWin64').once().returns(false);
+      let arch = await system.arch();
+      arch.should.equal('32');
+      mocks[SANDBOX].verify();
+  });
```
imurchie: Indentation.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
+
+
+});
```
imurchie: Indentation.

test/system-specs.js
```diff
@@ -11,13 +19,83 @@ describe('system', function () {
...
+
+
+  
+
 });
```
imurchie: Remove the extra blank lines.

test/system-specs.js
```diff
@@ -1,23 +1,88 @@
...
+
+    it('should correctly return Mac if it is a Mac', function () { 
+      let osMock = sinon.mock(os);
+      osMock.expects('type').returns('Darwin');
+      system.isMac().should.equal.true;
+      osMock.verify();
+      osMock.restore();
+    });
+
+    it('should correctly return Linus if it is a Linux', function () {
```
imurchie: Linux.

PR #13, opened 2015-07-10
Architecture detection support for windows 64/32

juanpicado:
I think this solve this bug https://github.com/appium/appium/issues/5289
Test included as well.

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -16,12 +16,12 @@ exports.isLinux = function () {
 };
 
 function isOSWin64() {
-    return process.arch === 'x64' || process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
+  return process.arch === 'x64' || process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
 }
 
 exports.arch = function (cb) {
   if (exports.isWindows() ) {
-      cb(null, isOSWin64() ? "64" : "32");
+    cb(null, isOSWin64() ? "64" : "32");
   } else {
     exec("uname -m", function (err, stdout) {
       if (err) return cb(err);
--- a/lib/system.js
+++ b/lib/system.js
@@ -15,13 +15,21 @@ exports.isLinux = function () {
   return !exports.isWindows() && !exports.isMac();
 };
 
+function isOSWin64() {
+    return process.arch === 'x64' || process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
+}
+
 exports.arch = function (cb) {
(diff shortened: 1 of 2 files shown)


jlipps:
thanks @juanpicado! I believe @soniayang97 is also fixing this bug in a different branch. we'll probably just publish that branch with the fix but i'll let her and @imurchie chime in

juanpicado:
yw, btw i forgot give the merit to @sebv , he posted a workaround.

imurchie:
Yes, a functionally equivalent change is [in the works](https://github.com/syang97/appium-support/blob/sonia-rewrite/lib/system.js).

If we are going to release another 1.4 patch, we could also merge this and release it in conjunction. Or we could wire in @soniayang97's  updated version when it has been done, which I think ought to be immanent (just waiting on a couple of tests).

Jonahss:
Hm. I want to do a 1.4.8 release. Should @soniayang97's updated version just go in to that? How about Tuesday (tomorrow) for that release?

Anything else we might want in that release? @imurchie @sebv @jlipps

Jonahss:
@moizjv wants it in.
After @soniayang97's work is merged, can you create a PR to appium to update the package.json. Then we can cherry-pick that into 1.4

lib/system.js
```diff
@@ -15,13 +15,21 @@ exports.isLinux = function () {
   return !exports.isWindows() && !exports.isMac();
 };
 
+function isOSWin64() {
+    return process.arch === 'x64' || process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
```
syang97: should be two spaces

lib/system.js
```diff
@@ -15,13 +15,21 @@ exports.isLinux = function () {
...
-  exec("uname -m", function (err, stdout) {
-    if (err) return cb(err);
-    if (stdout.trim() === "i686") {
-      cb(null, "32");
-    } else {
-      cb(null, "64");
-    }
-  });
+  if (exports.isWindows() ) {
+      cb(null, isOSWin64() ? "64" : "32");
```
syang97: indentation

syang97:
Besides the indentation, it pretty much looks like mine so maybe you can merge this first?

moizjv:
@juanpicado can you fix indentation so we can totally include this for 1.4.8 :)

juanpicado:
Sure ... In progress ..

Jonahss:
Great!
I squashed the "fixed indentation" commit into the previous commite and merged manually :)

PR #14, opened 2015-07-14
translate to es6/7

syang97:


lib/tempdir.js
```diff
@@ -1,45 +1,59 @@
...
+  let tempDirectory = await tempDir();
+  return nodePath.join(tempDirectory, name);
+}
 
+let open = async function (affixes) {
+  let filePath = await  path(affixes, 'f-'); 
+  try {
+    let fd = await fs.open(filePath, RDWR_EXCL, 0o600);
+    // opens the file in mode 384
+    return {path: filePath, fd: fd};
```
jlipps: this could be written slightly more compactly as:

```js
return {path: filePath, fd}
```

(when the key and the value have the same name, you don't need to specify both)

jlipps:
i made one tiny comment, then it looks good to merge to me!

jlipps:
awesome, merged!

PR #18, opened 2015-08-12
Add method to fs to parse plist

imurchie:
Add `parsePlist` method to `fs` object. I've found myself needing to promisify the plist parser again and again, so it needs to be somewhere. It is a file system method, so it seemed to make sense to be here.


lib/fs.js
```diff
@@ -1,9 +1,11 @@
 import _fs from 'fs';
 import rimraf from './rimraf';
 import { mkdir } from './util';
-import _md5 from 'MD5';
+import _md5 from 'md5';
```
jlipps: weird, was this not working before?

lib/fs.js
```diff
@@ -18,6 +20,7 @@ let fs = {
   rimraf,
   mkdir,
   copyFile: B.promisify(_ncp),
+  parsePlist: B.promisify(bplistParser.parseFile)
```
jlipps: i think @sebv is going to fight you over this. he didn't like it when i added non-node-fs methods to our fs bucket. i kind of disagree still :-)

jlipps:
apart from the larger issue that sebv and i were discussing, this PR LGTM :+1:

lib/fs.js
imurchie: Seems like I broke it before (when updating `md5` to version 2), and didn't notice. On my local machine it doesn't cause a problem, but it does on the Travis build.

imurchie:
Decided to add plist creation, at which time it made sense to move it into its own bucket.

sebv:
Yeah separate plist is better.

PR #21, opened 2015-08-17
adds mkdirp multiresolve

sebv:
ping @jlipps

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -2,6 +2,7 @@ import B from 'bluebird';
 import _ from 'lodash';
 import os from 'os';
 import fs from './fs';
+import path from 'path';
 
 export function hasContent (val) {
   return _.isString(val) && val !== "";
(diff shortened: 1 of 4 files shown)


lib/util.js
```diff
@@ -73,4 +74,14 @@ function cancellableDelay (ms) {
   });
 }
 
-export { hasValue, escapeSpace, hasAccess, mkdir, localIp, cancellableDelay };
+function multiResolve (roots) {
+  var args = Array.prototype.slice.call(arguments, 1);
+  var paths = [];
+  _.each(roots, function (root) {
```
jlipps: `for (let root of roots) {`

lib/util.js
```diff
@@ -73,4 +74,14 @@ function cancellableDelay (ms) {
   });
 }
 
-export { hasValue, escapeSpace, hasAccess, mkdir, localIp, cancellableDelay };
+function multiResolve (roots) {
```
jlipps: `function multiResolve (roots, ...args) {`

lib/util.js
```diff
@@ -73,4 +74,14 @@ function cancellableDelay (ms) {
...
+  var args = Array.prototype.slice.call(arguments, 1);
+  var paths = [];
+  _.each(roots, function (root) {
+    var resolveArgs = [root].concat(args);
+    paths.push(path.resolve.apply(null, resolveArgs));
```
jlipps: `paths.push(path.resolve(...resolveArgs))`

lib/util.js
jlipps: actually this is probably cleaner as `return roots.map(root => {`

PR #22, opened 2015-08-17
Fix wrong import of fs

imurchie:
Mixed up the import. Added a test so that it is caught.

### DIFF
--- a/test/plist-specs.js
+++ b/test/plist-specs.js
@@ -1,14 +1,31 @@
 import chai from 'chai';
 import path from 'path';
-import { plist } from '../..';
+import { plist, tempDir, fs } from '../..';
 
 
 chai.should();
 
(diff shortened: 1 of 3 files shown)


test/plist-specs.js
```diff
@@ -1,14 +1,31 @@
...
+
 describe('plist', () => {
-  it('parsePlistFile as binary', async () => {
-    let plistPath = path.resolve(__dirname, '..', '..', 'test', 'assets', 'sample.plist');
+  it('should parse plist file as binary', async () => {
     let content = await plist.parsePlistFile(plistPath);
     content.should.have.property('com.apple.locationd.bundle-/System/Library/PrivateFrameworks/Parsec.framework');
   });
+
+  it.only('should write plist file as binary', async () => {
```
jlipps: probably don't want the `.only` here, right?

PR #26, opened 2015-10-09
Fix mkdirp

imurchie:
Async `mkdirp` wasn't working. Manually promisify.


jlipps:
you know i think it was quasi working but only if you did a nextTick afterwards. this is good. thanks.

PR #28, opened 2015-10-14
Moved macOsxVersion to appium-support + bump to 2.0.4

scottdixon:
appium-doctor todo https://github.com/appium/appium-doctor/issues/3

@imurchie

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -27,4 +27,17 @@ async function arch () {
   }
 }
 
-export { isWindows, isMac, isLinux, isOSWin64, arch };
+async function macOsxVersion () {
+  let stdout;
+  try {
+    stdout = (await exec('sw_vers', ['-productVersion'])).stdout;
+  } catch (err) {
+    throw new Error(`Could not detect Mac OS X Version: ${err}`);
+  }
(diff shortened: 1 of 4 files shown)


lib/system.js
```diff
@@ -27,4 +27,17 @@ async function arch () {
   }
 }
 
-export { isWindows, isMac, isLinux, isOSWin64, arch };
+async function macOsxVersion () {
+  let stdout;
+  try {
+    stdout = (await exec('sw_vers', ['-productVersion'])).stdout;
+  } catch (err) {
+    throw new Error("Unknown SW Version Command: " + err);
```
imurchie: Can we make a more intuitive error message? Like "Unable to detect Mac OS X version: XXX" or something?

scottdixon: Good call. Line 40 should now take care of this - is there anything helpful we can bring out of line 35's error?

imurchie: I think it would still be good to have this error thrown, since there is an error object to give the details of.

scottdixon: I agree - but in regards to your 'intuitive' comment?

imurchie: I would just use

```js
throw new Error(`Could not detect Mac OS X Version: ${err}`);
```

PR #29, opened 2015-10-15
Add safeJson method to util

imurchie:


lib/util.js
```diff
@@ -86,4 +86,13 @@ function multiResolve (roots, ...args) {
   });
 }
 
-export { hasValue, escapeSpace, escapeSpecialChars, localIp, cancellableDelay, multiResolve };
+function safeJson (obj) {
```
jlipps: wondering if this is ambiguous between parse and stringify.

imurchie: That's a good point.

imurchie: https://github.com/appium/appium-support/pull/30

PR #31, opened 2016-01-07
Added mustExist flag to parsePlistFile

scottdixon:
In some cases, we don't want an error thrown if a plist file doesn't exist.

Please review @jlipps

### DIFF
--- a/lib/plist.js
+++ b/lib/plist.js
@@ -15,9 +15,19 @@ async function parseXmlPlistFile (plistFilename) {
   return  xmlplist.parse(xmlContent);
 }
 
-async function parsePlistFile (plist) {
+async function parsePlistFile (plist, mustExist = true) {
   log.debug(`Attempting to parse plist file '${plist}'`);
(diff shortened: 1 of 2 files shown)


lib/plist.js
```diff
@@ -43,7 +53,7 @@ async function updatePlistFile (plist, updatedFields, binary = true) {
   log.debug(`Attempting to update plist file'${plist}'`);
   let obj;
   try {
-    obj = await parsePlistFile(plist);
+    obj = await parsePlistFile(plist, false);
```
jlipps: this technically is making a breaking change to the behavior of this function. i think i would prefer to punt on the choice to the user in this case, adding another `mustExist = true` param and simply passing it on, so that we're not making a breaking change to this library

jlipps:
otherwise LGTM!

lib/plist.js
imurchie: I tend to agree.

PR #35, opened 2016-03-25
Use map instead of pluck

imurchie:
Lodash stopped having a `pluck` method.

Also, fix the test for `util.localIp` so that it actually tests something, rather than just mocking the method under test. :(


imurchie:
This should fix https://github.com/appium/appium/issues/6335

PR #36, opened 2016-04-06
Use streaming md5 library

imurchie:
Otherwise large files will throw an error.

```
FATAL ERROR: invalid table size Allocation failed - process out of memory
```


imurchie:
The `md5-file` library [purports to be able to hash a 4Gb file](https://github.com/roryrjb/md5-file#md5-file--). I have successfully hashed a 2Gb apk, which is the largest I could build with `ant`.

test/fs-specs.js
```diff
@@ -73,9 +73,15 @@ describe('fs', function () {
...
-  it('md5', async () => {
-    let existingPath = path.resolve(__dirname, 'fs-specs.js');
-    (await fs.md5(existingPath)).should.have.length(32);
+  describe('md5', () => {
+    it('should calculate hash of correct length', async () => {
+      let existingPath = path.resolve(__dirname, 'fs-specs.js');
+      (await fs.md5(existingPath)).should.have.length(32);
+    });
+    it('should be able to run on huge file', async () => {
+      let existingPath = path.resolve(__dirname, '..', '..', 'test', 'assets', 'enormous.apk');
```
jlipps: how big is enormous.apk? do we really want to add it to git? could we also simply generate a huge file that is a list of 1s or something so we know the md5 beforehand and can test it?

test/util-specs.js
```diff
@@ -115,7 +115,7 @@ describe('util', function () {
     });
   });
 
-  describe.only('localIp', function () {
+  describe('localIp', function () {
```
jlipps: eek!

jlipps:
one comment then LGTM

PR #37, opened 2016-04-10
add fs.read function to fs promisified functions

cstyan:
node-adb-client makes use of `fs.read` to read specific chunk sizes from a file


imurchie:
Cool! Thanks!

imurchie:
Published as `appium-support@2.2.1`

PR #39, opened 2016-04-27
Await on the returned Promise from mkdir

alewang:
Avoids race condition where code run later expects there to be a tmp directory prepared.


imurchie:
Good catch! This is the drawback of the async/await code, though it beats the drawbacks of callbacks!

Thanks!

imurchie:
Published as `appium-support@2.3.0`.

PR #40, opened 2016-06-22
Allow macOS 10.12 to be found

imurchie:
Add latest macOS release to list of acceptable versions. While this will not necessarily mean that macOS 10.12 is supported, it at least means that it will not be stopped by simple checks for validity.

Addresses https://github.com/appium/appium/issues/6609


jlipps:
👍

PR #41, opened 2016-09-07
Add unwrapElement method to util

imurchie:
We use this in ios and xcuitest, and probably elsewhere in _ad hoc_ ways. It's simply, but does not need to be repeated.


jlipps:
👍

PR #44, opened 2016-12-17
Move logger from appium-base-driver into support

imurchie:
Move logger into the support package, for use from other packages. @jlipps


imurchie:
Published in `appium-support@2.4.0`

PR #47, opened 2017-03-08
Add helpers for process management

imurchie:
In a few places we check for a process, and kill it. Consolidate it so that it is tested properly and available to whomever needs it.


imurchie:
I think `teen_process` is conceptually a replacement for `child_process`, and these don't really conceptually fit into that module.

jlipps:
sure, that makes sense!

PR #48, opened 2017-04-05
Decrease maximum log records count to stabilise memory usage

mykola-mokhnach:
Reduce memory usage by limiting the size of internal log buffer.
records


mykola-mokhnach:
it looks like it: https://github.com/npm/npmlog/blob/master/log.js. Check the _log.record_ array.
Also, they have _log._buffer_ property, which is populated when logging is paused and where the size is not limited at all. So be careful when using pause/resume methods for logger.

imurchie:
Published in `appium-support@2.7.0`

PR #49, opened 2017-04-21
Add zip methods to appium support

dpgraham:
* Uses `extract-zip` and `yauzl` (yauzl is dependency of extract-zip) to extract zip files
* Uses `archiver` to create zip files
* Added methods `extractAllTo`, `readEntries` and `toInMemoryZip` to the API
* Next step is to refactor out other repos that use `adm-zip` and use these methods instead

(https://github.com/appium/appium/issues/7409)


lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {str} destDir 
+ */
+async function _extractEntryTo (zipfile, entry, destDir) {
+  // Create directory if doesn't exist already
+  const relDir = path.dirname(entry.fileName);
+  await mkdirp(path.resolve(destDir, relDir));
+  
+  // Stream the contents of the entry to a file
+  return new B((resolve) => {
+    zipfile.openReadStream(entry, (err, readStream) => {
```
imurchie: Do we not want to check the error here, and `reject` the promise if necessary?

lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+      const writeStream = createWriteStream(path.resolve(destDir, entry.fileName), {flags: 'w'});  
+      readStream.pipe(writeStream);
+      readStream.on('end', resolve);
```
imurchie: We might also want error handling on the stream.

lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+async function toInMemoryZip (srcDir) {
+  const tempOutputDir = await tempDir.openDir();
+  return new B((resolve, reject) => {
+    // Create a writable stream that zip is streamed to 
+    const output = createWriteStream(path.resolve(tempOutputDir, 'temp.zip'));
+    output.on('finish', async () => {
+      try {
+        const zipContents = await fs.readFile(path.resolve(tempOutputDir, 'temp.zip'));
+        resolve(zipContents);
+      } catch (err) { reject(err); }
```
imurchie: Put the `reject` on a new line.

lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    // Start streaming zip data to the above writable stream
+    const archive = archiver('zip', {
+      zlib: { level: 9 }
```
imurchie: No spaces around object literal.

lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {Object} entry 
+ * @param {str} destDir 
+ */
+async function _extractEntryTo (zipfile, entry, destDir) {
+  // Create directory if doesn't exist already
+  const relDir = path.dirname(entry.fileName);
+  await mkdirp(path.resolve(destDir, relDir));
+  
+  // Stream the contents of the entry to a file
+  return new B((resolve) => {
```
imurchie: `return await new B((resolve) => {`

lib/zip.js
```diff
@@ -0,0 +1,93 @@
...
+  });
+}
+
+/**
+ * Converts contents of local directory to an in-memory .zip buffer
+ * @param {*} srcDir 
+ */
+async function toInMemoryZip (srcDir) {
+  const tempOutputDir = await tempDir.openDir();
+  return new B((resolve, reject) => {
```
imurchie: `return await new B((resolve, reject) => {`

dpgraham:
@imurchie I agree with the confusion of mixing promises and async. I overhauled it to use mostly async and only minimally use promises in cases where we need to bind to EventEmitter events.

e.g.:
```javascript
let promise = new B((resolve, reject) => {
  eventEmitter.on('finished', resolve);
  eventEmitter.on('error', reject);
});
```

I also updated the tests like you asked to force more errors. Good call on that, I spotted some issues with the code when I did that.

imurchie:
Testing FTW!

lib/zip.js
```diff
@@ -0,0 +1,116 @@
...
+  const writeStream = createWriteStream(path.resolve(destDir, entry.fileName), {flags: 'w'});  
+  const writeStreamPromise = new B((resolve, reject) => {
+    writeStream.once('finish', resolve);
+    writeStream.once('error', reject);
+  });
+
+  // Create zipReadStream and pipe data to the write stream
+  // (for some odd reason B.promisify doesn't work on zipfile.openReadStream, it causes an error 'closed')
+  const zipReadStream = await new B((resolve, reject) => {
+    zipfile.openReadStream(entry, (err, readStream) => err ? reject(err) : resolve(readStream));
```
imurchie: Slick!

lib/zip.js
```diff
@@ -0,0 +1,116 @@
...
+  const zipWriteStreamPromise = new B((resolve, reject) => {
+    zipWriteStream.once('finish', resolve);
+    zipWriteStream.once('error', reject);
+  });
+
+  // Zip 'srcDir' and stream it to the above writable stream
+  const archive = archiver('zip', {
+    zlib: {level: 9}    
+  });
+  archive.once('error', (err) => { throw new Error(err); });
```
imurchie: `err` ought to already be an `Error` object, so re-wrapping it is not necessary.

test/zip-specs.js
```diff
@@ -0,0 +1,115 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import path from 'path';
+import { zip } from '..';
+import { tempDir, fs } from '../index';
```
imurchie: These two are the same import, no? `import { zip, tempDir, fs } from '..';` would suffice.

imurchie:
Published in `appium-support@2.8.0`.

PR #50, opened 2017-04-25
Changes to zip method and zip testing

dpgraham:
* Refactored the `zip.toInMemoryZip` to not use temporary directory as writeStream but instead uses an ad-hoc writeStream that writes the zip buffers to an array and then concatenates them together at the end (In an earlier PR Isaac pointed out that file write streams were being used unnecessarily and it occurred to me that the file stream was unnecessary in this method)
* Refactored `zip-specs.js` to use mock filesystem instead of the actual filesystem
* Updated the documentation to include the zip methods


imurchie:
@dpgraham Is this still on the go?

dpgraham:
I still need to commit this. Let me fix the conflicts and then I'll merge it.

imurchie:
👍

dpgraham:
Published to appium-support@2.8.2

PR #51, opened 2017-04-28
Fail zipping directory to which we have no access

imurchie:
My editor deletes trailing spaces, so there is some chaff in there. The essential part is `lib/zip.js`, lines 90-92. If the directory is not there, fail.

Otherwise the `appium-xcuitest-driver` [build fails catastrophically](https://travis-ci.org/appium/appium-xcuitest-driver/jobs/226870050#L5961-L5982) on [this test](https://github.com/appium/appium-xcuitest-driver/blob/master/test/functional/basic/file-movement-e2e-specs.js#L59-L61).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 build
 coverage
 *.log
+.DS_Store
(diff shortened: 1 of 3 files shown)


dpgraham:
Looks good to me.

dpgraham:
Published to: appium-support@2.8.1

PR #53, opened 2017-09-29
Add a possibility to pass dynamic prefix generator to a logger instance

mykola-mokhnach:
This will be needed to implement more advanced logging mechanism for parallel drivers. The idea is to create a single logger instance per driver instance and, as a log prefix, provide a special function, which adds session id if it exists for the current driver instance.

P. S. The whole idea might suck, so it would be pretty ok if this PR is closed :)

### DIFF
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -1,4 +1,5 @@
 import npmlog from 'npmlog';
+import _ from 'lodash';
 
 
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@imurchie added tests

imurchie:
Published in `appium-support@2.9.0`.

PR #54, opened 2017-10-10
Add mac 10.13 to list of possible mac versions

imurchie:
Fixes https://github.com/appium/appium/issues/9415

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -34,7 +34,7 @@ async function macOsxVersion () {
   } catch (err) {
     throw new Error(`Could not detect Mac OS X Version: ${err}`);
   }
-  for (let v of ['10.8', '10.9', '10.10', '10.11', '10.12']) {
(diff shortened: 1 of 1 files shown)


lib/system.js
```diff
@@ -34,7 +34,7 @@ async function macOsxVersion () {
   } catch (err) {
     throw new Error(`Could not detect Mac OS X Version: ${err}`);
   }
-  for (let v of ['10.8', '10.9', '10.10', '10.11', '10.12']) {
+  for (let v of ['10.8', '10.9', '10.10', '10.11', '10.12', '10.13']) {
```
mykola-mokhnach: Can we move this array into a constant? I think it might be more readable to rewrite it like this:

```js
if (const ver = _.find(SUPPORTED_MAC_OS_VERSIONS, (v) => stdout.startsWith(v))) {
    return ver;
}
throw ...
```

imurchie: Updated.

PR #55, opened 2017-10-27
Add leak-proof truncate function

dpgraham:
* Calls _.truncate but does a hack of adding (' ' + _.truncate(...args)).substr(1) that creates a new string instead of keeping it in memory
* This is due to a bug/feature in V8 (https://bugs.chromium.org/p/v8/issues/detail?id=2869)
* String.substr doesn't truncate the string, it keeps the whole string in memory
error: pathspec ' + _.truncate(...args)).substr(1) that creates a new string instead of keeping it in memory
* This is due to a bug/feature in V8 (https://bugs.chromium.org/p/v8/issues/detail?id=2869)
* String.substr doesnt' did not match any file(s) known to git

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -109,5 +109,13 @@ function unwrapElement (el) {
   return el;
 }
 
+/*
+ * Works the same way as _.truncate, but prevents memory leaks (see https://bugs.chromium.org/p/v8/issues/detail?id=2869) 
+ */
+function truncate (...args) {
+  // Hack that doesn't keep the whole string in memory
+  return ((' ') + _.truncate(...args)).substr(1);
+}
+
 export { hasValue, escapeSpace, escapeSpecialChars, localIp, cancellableDelay,
-         multiResolve, safeJsonParse, unwrapElement };
+         multiResolve, safeJsonParse, unwrapElement, truncate };
--- a/lib/util.js
+++ b/lib/util.js
@@ -110,7 +110,7 @@ function unwrapElement (el) {
 }
 
 /*
- * Works the same way as _.truncate, but prevents memory leaks (see https://bugs.chromium.org/p/v8/issues/detail?id=2869) 
+ * Works the same way as _.truncate, but prevents memory leaks (see https://bugs.chromium.org/p/v8/issues/detail?id=2869)
  */
 function truncate (...args) {
   // Hack that doesn't keep the whole string in memory
--- a/test/e2e/util-e2e-specs.js
+++ b/test/e2e/util-e2e-specs.js
@@ -1,0 +1,19 @@
+
+import { util } from '../..';
+import logger from '../../lib/logger';
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+chai.use(chaiAsPromised);
+
+describe('util', function () {
+  describe('truncate', () => {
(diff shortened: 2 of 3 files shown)


dpgraham:
Related to https://github.com/appium/appium/issues/9543

mykola-mokhnach:
This function is good, but we can also apply it directly to our logger method inside this module,so it's not needed to change anything in all the client libraries.

imurchie:
I think we'd still need to update the client libraries, since they use `_.truncate` directly.

mykola-mokhnach:
This will require to change all the drivers and other helper modules, which use logging. I assume it would be much easier to add this workaround into our logger wrapper here: https://github.com/appium/appium-support/blob/1245ba85dea79a7c87e66064725440c6cbceb548/lib/logging.js#L60

mykola-mokhnach:
The problem there is the fact that the default logger writes logged valus into an internal buffer, which prevents them from being garbage collected. If we apply the workaround on this stage then correct string copies will be recorded into the buffer and their non-truncated "parents" will be successfully garbage-collected.

imurchie:
That would be good, but would only solve the problem when we send big values to be logged. In most places the calling code truncates it and passes the shortened version along. So _those_ places will still need to be fixed.

mykola-mokhnach:
>  In most places the calling code truncates it and passes the shortened version along. So those places will still need to be fixed.

Do you mean the generic case? Because my intension was to address the particular logger memory leak issue, which is mentioned above, in the first priority.

imurchie:
Maybe I'm misunderstanding. However, I agree it should be fixed in the logger. 

But if `_.truncate` has the problem too, I don't think fixing the logger will actually have much affect on things.

mykola-mokhnach:
> But if _.truncate has the problem too, I don't think fixing the logger will actually have much affect on things.

It will. The size of logging buffer equals to 3000 records for now. The maximum size of NodeJS heap is limited to ~1GB (actually, is different for x86/x64 and also depends on the version). This means it's quite easy to expire the free space in the heap by writing big strings there (1000 / 3000 ~ 0.33 MB max per single log record, or even less in the real world, since the heap is also containing many other objects)

imurchie:
Ok

dpgraham:
@mykola-mokhnach I agree that it's better to do it in the actual debugger instead of augmenting `_.truncate`; I'll make that change

mykola-mokhnach:
@dpgraham What do you think about adding a PR to lodash module, which workarounds this behaviour?

dpgraham:
@mykola-mokhnach I thought about that. I think the downside of the patch fix is the performance issues given that `(' ' + str).substr(1)` creates an entirely new string and adds it to memory. 

Perhaps a PR that does a size difference check:

e.g.)
```javascript
_.truncate = function(str, {length}) {
  str result = /* truncate method */;
  if (str.length - length > 100) {
    return (' ' + result).substr(1);
  }
}
```

mykola-mokhnach:
@dpgraham yes, makes sense

mykola-mokhnach:
@dpgraham Do we still need to merge this PR or it can be closed?

dpgraham:
@mykola-mokhnach We can close this. The fixes you made made this obsolete.

PR #56, opened 2017-11-01
Workaround memory leak issue caused by truncated strings

mykola-mokhnach:
This should address memory leak issue caused by the fact that original strings are kept in memory after _.truncate operation is done on them. And since logs have an internal buffer of size 3000, where all logged strings are stored by by default, this might cause serious leaks in case some really big strings are being truncated and then passed to logs.


imurchie:
Published in `appium-support@2.11.0`.

PR #57, opened 2017-12-07
Fix tests on Mac 10.13

imurchie:
Some general cleanup. Main change is to use small size for blocks to write in md5 test. On 10.13 the blocksize is huge and it slows the test down (or causes a memory error). 4096 bytes is ample and the tests remain fast.

### DIFF
--- a/test/fs-specs.js
+++ b/test/fs-specs.js
@@ -7,6 +7,7 @@ import { exec } from 'teen_process';
 let should = chai.should();
 
 describe('fs', function () {
+  const existingPath = path.resolve(__dirname, 'fs-specs.js');
   it("should exist", function () {
     should.exist(fs);
   });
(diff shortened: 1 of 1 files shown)


test/fs-specs.js
```diff
@@ -45,81 +46,76 @@ describe('fs', function () {
...
   });
-  it('stat', async () => {
-    let existingPath = path.resolve(__dirname, 'fs-specs.js');
+  it('stat', async function () {
     let stat = await fs.stat(existingPath);
     stat.should.have.property('atime');
   });
-  describe('which', () => {
+  describe('which', function () {
     it('should find correct executable', async () => {
```
mykola-mokhnach: async function ?

imurchie: Yup.

PR #58, opened 2017-12-29
Add image-util with methods for cropping images

astappev:
See [this](https://github.com/appium/appium-base-driver/pull/144) and [this thread](https://github.com/appium/appium-base-driver/pull/148) for more details

PR #59, opened 2017-12-29
add image helpers

jlipps:
superseding https://github.com/appium/appium-support/pull/58, disabling eslint is not required. once this passes CI it can be merged. cc @mykola-mokhnach


lib/image-util.js
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Create a pngjs image from given base64 image
+ *
+ * @param {string} base64Image The string with base64 encoded image
+ * @return {PNG} The image object
+ */
+async function base64ToImage (base64Image) {
+  const imageBuffer = new Buffer(base64Image, 'base64');
+  // const image = new PNG({filterType: 4});
```
mykola-mokhnach: do we need these comments?

lib/image-util.js
```diff
@@ -0,0 +1,116 @@
...
+  // const parse = B.promisify(image.parse, {context: image});
+  // return await parse(imageBuffer);
+
+  return new B((resolve, reject) => {
+    const image = new PNG({filterType: 4});
```
mykola-mokhnach: is there a constant name for this particular filter type?

lib/image-util.js
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Create a base64 string for given image object
+ *
+ * @param {PNG} image The image object
+ * @return {string} The string with base64 encoded image
+ */
+async function imageToBase64 (image) {
+  return new B((resolve, reject) => {
+    const chunks = [];
+    image.pack().on('data', (chunk) => chunks.push(chunk)).on('end', () => {
```
mykola-mokhnach: I'd rather move `.on` to the next line for better reading

lib/image-util.js
```diff
@@ -0,0 +1,116 @@
...
+  for (let y = firstVerticalPixel; y < lastVerticalPixel; y++) {
+    for (let x = firstHorizontalPixel; x < lastHorizontalPixel; x++) {
+      const firstByteIdxInPixelBlock = (imageRect.width * y + x) << 2;
+      for (let byteIdx = 0; byteIdx < BYTES_IN_PIXEL_BLOCK; byteIdx++) {
+        croppedArray.push(image.data[firstByteIdxInPixelBlock + byteIdx]);
+      }
+    }
+  }
+
+  image.data = new Buffer(croppedArray);
```
mykola-mokhnach: is that expected that the original image is mutated?

lib/image-util.js
```diff
@@ -0,0 +1,116 @@
...
+  image.width = interRect.width;
+  image.height = interRect.height;
+  return image;
+}
+
+function getRectIntersection (rect, imageSize) {
```
mykola-mokhnach: I'm not quite sure this algorithm is correct.
This is the implementation of the same method from java.awt lib:

```java
   public Rectangle intersection(Rectangle r) {
        int tx1 = this.x;
        int ty1 = this.y;
        int rx1 = r.x;
        int ry1 = r.y;
        long tx2 = tx1; tx2 += this.width;
        long ty2 = ty1; ty2 += this.height;
        long rx2 = rx1; rx2 += r.width;
        long ry2 = ry1; ry2 += r.height;
        if (tx1 < rx1) tx1 = rx1;
        if (ty1 < ry1) ty1 = ry1;
        if (tx2 > rx2) tx2 = rx2;
        if (ty2 > ry2) ty2 = ry2;
        tx2 -= tx1;
        ty2 -= ty1;
        // tx2,ty2 will never overflow (they will never be
        // larger than the smallest of the two source w,h)
        // they might underflow, though...
        if (tx2 < Integer.MIN_VALUE) tx2 = Integer.MIN_VALUE;
        if (ty2 < Integer.MIN_VALUE) ty2 = Integer.MIN_VALUE;
        return new Rectangle(tx1, ty1, (int) tx2, (int) ty2);
    }

```

jlipps:
@mykola-mokhnach addressed all your remarks with changes or comments in the code

lib/image-util.js
```diff
@@ -0,0 +1,118 @@
...
+  const left = rect.left >= imageSize.width ? imageSize.width : rect.left;
+  const top = rect.top >= imageSize.height ? imageSize.height : rect.top;
+
+  // if we can't find a total intersection of rect with imageSize, then say
+  // that width and/or height are 0, to indicate we didn't get a valid
+  // intersection
+  const width = imageSize.width >= (left + rect.width) ? rect.width : 0;
```
mykola-mokhnach: I'm afraid this may not work properly if the intersection is not full `(left < imageSize.width && left + rect.width > imageSize.width)`. The resulting `width` should not be zero in  such case

jlipps: I think the point is that we do not allow non-full intersections, and setting the width to 0 is a way of doing that. We could also throw an error instead?

lib/image-util.js
astappev: No, sorry, forgot to remove

lib/image-util.js
astappev: Yes @jlipps, you are right, we don't need to support partial intersections (for current use-case).
Also, we already have an error which will be thrown in such case (on line 77), but to be more clear, we can add addition one inside getRectIntersection.

mykola-mokhnach: Then we need to adjust the function name, otherwise it's misleading.

mykola-mokhnach: I'd expect partial intersections to be calculated properly, since it's a pretty valid case

jlipps: ok @mykola-mokhnach @astappev i changed things around so the intersection logic itself is more general and the calling function does the check and the exception to keep the current overall behavior. what do you think?

jlipps:
@mykola-mokhnach @astappev what do you think about the new rect intersection logic?

mykola-mokhnach:
@jlipps I'm happy

jlipps:
\o/

jlipps:
published in `appium-support@2.12.0`

PR #64, opened 2018-02-21
Fix protocol name comparsion

mykola-mokhnach:
It looks like the parsed protocol name also includes a colon at the end


jlipps:
i mean another way to solve this is just use the `startswith` check for ftp as well

mykola-mokhnach:
yes it was, it looks like nobody tried it so far :)

PR #65, opened 2018-04-13
Add OpenCV helpers for image comparison

mykola-mokhnach:
The actual implementation is based on [opencv4nodejs](https://github.com/justadudewhohacks/opencv4nodejs/blob/2f594c4739b550a6b6413a454fe14f1b7617e4a6/README.md) features: Template Matching and Feature Matching. The next step would be to create dedicated endpoints in the base driver and use this API there for advanced images/screenshots comparison.

I've disabled OpenCV build on package install (it takes more than 10 minutes), so the client will have to install all the necessary native libs manually if these helpers are needed (for example via `brew`).


lib/image-util.js
```diff
@@ -13,6 +15,88 @@ import B from 'bluebird';
...
+ * for brute-force matching.
+ * Read https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_feature2d/py_matcher/py_matcher.html
+ * for more details.
+ *
+ * @param {cv.FeatureDetector} detector OpenCV feature detector instance
+ * @param {Buffer} imgData image data packed into a NodeJS buffer
+ *
+ * @returns {cv.DescriptorMatch} OpenCV match descriptor
+ */
+async function detectAndCompute (detector, imgData) {
```
imurchie: In terms of an API, this feels off to me. `detectAndCompute` takes the image data second and detector first, while `getMatchesCount` takes the image data first and the detector second.

mykola-mokhnach: this is an internal helper, but I can change arguments order

lib/image-util.js
```diff
@@ -1,6 +1,8 @@
-import {Buffer} from 'buffer';
-import {PNG} from 'pngjs';
+import _ from 'lodash';
+import { Buffer } from 'buffer';
+import { PNG } from 'pngjs';
 import B from 'bluebird';
+import * as cv from 'opencv4nodejs';
```
jlipps: does this do something different from just `import cv from 'opencv4nodejs';`?

lib/image-util.js
```diff
@@ -13,6 +15,88 @@ import B from 'bluebird';
...
+ * @param {Buffer} imgData1 the data of the first image packed into a NodeJS buffer
+ * @param {Buffer} imgData2 the data of the second image packed into a NodeJS buffer
+ * @param {string} detectorName one of possible OpenCV feature detector names:
+ * AKAZE, AGAST, BRISK, FAST, GFTT, KAZE, MSER, ORB
+ *
+ * @returns {number} The count of detected matches between the two images. Zero is returned
+ * if no matches are found.
+ * @throws {Error} If `detectorName` value is unknown.
+ */
+async function getMatchesCount (imgData1, imgData2, detectorName) {
```
jlipps: how do you imagine this method will be used? or do you think `getSimilarityScore` is the primary method for consumers?

mykola-mokhnach: these two methods implement totally different approaches. `getMatchesCount` calculates common edges. this means such method might work for cases, where two images are rotated, resized or desaturated, but still have similar content. The second method does pixel-by-pixel comparison and only works for images of equal size. It might be ideal solution for cases, where we need to compare screenshots for the same screen area after some time period.

lib/image-util.js
mykola-mokhnach: the other one didn't work for me :(

mykola-mokhnach:
@imurchie Can you please help me to properly setup `package.js`.  opencv4nodejs's tutorial says one has to set `OPENCV4NODEJS_DISABLE_AUTOBUILD` environment variable in order to skip opencv building. I've added this variable to the `config` section of the package config, but it looks like this didn't help and Travis still tries to [build](https://travis-ci.org/appium/appium-support/builds/366245957?utm_source=github_status&utm_medium=notification) the whole lib :)

package.json
```diff
@@ -43,10 +43,14 @@
     "request-promise": "^4.2.2",
     "source-map-support": "^0.4.6",
     "teen_process": "^1.5.1",
+    "opencv4nodejs": "^4.2.0",
     "which": "^1.2.4",
     "yauzl": "^2.7.0",
     "pngjs": "^3.0.0"
   },
+  "config": {
```
jlipps: i think you actually have to set this as an env var outside of package.json. so in a travis env. this isn't ideal, but i don't think there's a way we can force an env var to be a certain value just during npm install. we might have to fork opencv4nodejs and flip the default switch. or add a custom script that only installs/builds opencv4 if an appium user passes a certain flag. hmm.

package.json
```diff
@@ -49,7 +49,8 @@
     "pngjs": "^3.0.0"
   },
   "scripts": {
-    "prepublish": "OPENCV4NODEJS_DISABLE_AUTOBUILD=1 gulp prepublish",
+    "prepublish": "gulp prepublish",
+    "install": "OPENCV4NODEJS_DISABLE_AUTOBUILD=1 node-gyp rebuild",
```
jlipps: good idea

mykola-mokhnach:
@KazuCocoa Do you have any proposals/suggestion to this feature?

lib/image-util.js
```diff
@@ -10,9 +12,142 @@ import B from 'bluebird';
...
+ * Calculates the occurence position of a partial image in the full
+ * image.
+ *
+ * @param {Buffer} fullImgData The data of the full image packed into a NodeJS buffer
+ * @param {Buffer} partialImgData The data of the partial image packed into a NodeJS buffer
+ *
+ * @returns {Point} The absolute coordinates of the partial image occurence
+ * @throws {Error} If no occurences of the partial image can be found in the full image
+ */
+async function getMatchPosition (fullImgData, partialImgData) {
```
jlipps: @mykola-mokhnach can we also return the bounding rect of the match? for example, if a match is found but at a different scale, then we will not know where to tap. it would be nice to get the x/y of the top left corner of the match, plus the width/height it matched at, if that is possible.

KazuCocoa:
As a platform like Appium, it's good way to provide OpenCV based feature to users since we need to cover many needs. In the case, OpenCV is choose for me, I believe.

One suggestion is what about provide returning image diff?

I've used https://github.com/yahoo/kobold and https://github.com/yahoo/blink-diff to comparing images since I needed only `png` image comparison and detecting comparison area. OpenCV was overkilled for the purpose then. But recently I investigated a bit OpenCV to make image comparison flexible.
What I like the blink-diff and kobold, we can get diff image like below. (right side's image). And we can set some filters like https://github.com/yahoo/blink-diff#cropping. They helped our error detections and investigate the failed cause.
we can generate similar image using opencv and so, I thought it's nice if we can provide the diff image as a response.

lib/image-util.js
```diff
@@ -10,9 +12,142 @@ import B from 'bluebird';
...
+ * It is expected, that both images have the same resolution.
+ *
+ * @param {Buffer} templateData The data of the first image packed into a NodeJS buffer
+ * @param {Buffer} referenceData The data of the second image packed into a NodeJS buffer
+ *
+ * @returns {number} The similarity score as a float number in range [0.0, 1.0].
+ * 1.0 is the highest score (means both images are totally equal).
+ * @throws {Error} If the given images have different resolution.
+ */
+async function getSimilarityScore (templateData, referenceData) {
```
KazuCocoa: 👍

KazuCocoa:
https://github.com/KazuCocoa/opencv_sample/blob/master/lib/opencv.rb#L15:10 and 
 https://github.com/KazuCocoa/opencv_sample/tree/master/images/results are examples to get image diff like my above comment using opencv, btw.

saikrishna321:
@mykola-mokhnach can we handling fuzz factor as well ?

KazuCocoa:
Wow, brilliant 👍 @mykola-mokhnach

lib/image-util.js
```diff
@@ -10,9 +12,252 @@ import B from 'bluebird';
...
+ */
+function initOpenCV () {
+  if (!cv) {
+    try {
+      cv = require('opencv4nodejs');
+    } catch (ign) {}
+  }
+  if (!cv) {
+    throw new Error('opencv4nodejs module is required to use OpenCV features. ' +
+                    'Please install it first (npm i -g opencv4nodejs) and restart node. ' +
```
jlipps: `and restart Appium`

lib/image-util.js
```diff
@@ -10,9 +12,252 @@ import B from 'bluebird';
...
+  initOpenCV();
+
+  const {detectorName='ORB', visualize=false} = options;
+  if (!_.includes(AVAILABLE_DETECTORS, detectorName)) {
+    throw new Error(`'${detectorName}' detector is unknown. ` +
+                    `Only ${JSON.stringify(AVAILABLE_DETECTORS)} detectors are supported.`);
+  }
+
+  const detector = new cv[`${detectorName}Detector`]();
+  const img1 = await cv.imdecodeAsync(imgData1);
```
jlipps: can these decodes be done in parallel?

lib/image-util.js
```diff
@@ -10,9 +12,252 @@ import B from 'bluebird';
...
+ * @param {Buffer} img2Data The data of the second image packed into a NodeJS buffer
+ * @param {?SimilarityOptions} options [{}] Set of similarity calculation options
+ *
+ * @returns {SimilarityResult} The calculation result
+ * @throws {Error} If the given images have different resolution.
+ */
+async function getImagesSimilarity (img1Data, img2Data, options = {}) {
+  initOpenCV();
+
+  const {visualize=false} = options;
```
jlipps: i'm not sure i want a different rule about assignment spacing when it's default assignment inside an object destructuring. i.e., i wonder if to keep things consistent we want this style:

```js
const {visualize = false} = options;
```

i'm open to debate here. @imurchie what do you think?

lib/image-util.js
```diff
@@ -10,9 +12,252 @@ import B from 'bluebird';
...
+  reference = await reference.convertToAsync(cv.CV_8UC3);
+
+  const matched = await reference.matchTemplateAsync(template, cv.TM_CCOEFF_NORMED);
+  const minMax = await matched.minMaxLocAsync();
+  const result = {
+    score: minMax.maxVal
+  };
+  if (visualize) {
+    const resultMat = new cv.Mat(template.rows, template.cols * 2, cv.CV_8UC3);
+    await reference.copyToAsync(
```
jlipps: can these 2 steps be run in parallel?

lib/image-util.js
mykola-mokhnach: done

lib/image-util.js
mykola-mokhnach: not a problem at all

lib/image-util.js
mykola-mokhnach: yep

mykola-mokhnach:
@jlipps 

> i still have the question about returning the rect for the matched partial though.

Unfortunately, I haven't found any possibility to match resized rect. OpenCV can only find the left top corner and the match size must be the same as the template size:

https://docs.opencv.org/2.4/doc/tutorials/imgproc/histograms/template_matching/template_matching.html

mykola-mokhnach:
I think your proposal can be theoretically achieved via feature matching strategy, since it can match rotated and resized pictures

mykola-mokhnach:
@jlipps Added bounding rect calculation to feature matching algorithm. The resulting visualization with good matches factor 40 looks like:

result

KazuCocoa:
🆒

mykola-mokhnach:
Occurrences lookup result example:
example

PR #66, opened 2018-04-19
Return full size of the matched area

mykola-mokhnach:
I think this might be more handy for the clients

### DIFF
--- a/lib/image-util.js
+++ b/lib/image-util.js
@@ -360,7 +360,7 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
 
 /**
  * @typedef {Object} OccurrenceResult
- * @property {Rect} location The absolute location of the partial image occurence
+ * @property {Rect} rect The region of the partial image occurence
  * on the full image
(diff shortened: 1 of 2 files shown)


lib/image-util.js
```diff
@@ -360,7 +360,8 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
 
 /**
  * @typedef {Object} OccurrenceResult
- * @property {Point} location The absolute coordinates of the partial image occurence
+ * @property {Rect} location The absolute location of the partial image occurence
```
imurchie: This seems like a confusing name, now, since it is not a "location" if it includes a width and height.

mykola-mokhnach: which one would be better?

imurchie: Well, throughout we (and Selenium) have used `rect` as the thing which has a position and dimensions.

mykola-mokhnach: ok, changed to `rect`

imurchie:
Published in `appium-support@2.15.0`.

PR #69, opened 2018-05-01
Fix uploading of files

imurchie:
Fix importing and using of `fs` module.

Fixes https://github.com/appium/appium/issues/10630

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ build
 coverage
 *.log
 .DS_Store
-package-lock.json
+package-lock.json*
--- a/lib/net.js
+++ b/lib/net.js
@@ -1,5 +1,5 @@
 import { createReadStream } from 'fs';
-import { stat, exists } from './fs';
+import fs from './fs';
 import url from 'url';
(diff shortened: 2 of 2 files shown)


mykola-mokhnach:
Thanks

hursevich:
Thanks, guys for quick fix! Waiting for next release to check, not sure how the ticket resolution flow is working in Appium development?

jlipps:
@hursevich this has not yet been merged. when it is, the package will then need to be published, at which point the update is usually available by reinstalling the beta version of appium via npm.

imurchie:
It fixes it in that `exists` is called in the context of `fs`, rather than nothing. `fs.exists` is just a helper, calling `this.hasAccess` (https://github.com/appium/appium-support/blob/master/lib/fs.js#L23), which is not there when the function is called alone, without binding first.

imurchie:
Published in `appium-support@2.16.1`.

jlipps:
ahh, it was a binding issue. ok 👍

PR #74, opened 2018-06-12
reject image template matches below a threshold of default 0.5

jlipps:
my belief is that the primary use of this feature is to find things that exist on the screen, not to find the most likely place for them to exist (even when they don't).

to get the old behavior, just set threshold to 0


mykola-mokhnach:
It might be handy to have a test for this feature

jlipps:
published as 2.17.0

PR #75, opened 2018-07-05
add mjpeg stream helper class

jlipps:
the new version of https://github.com/appium/appium-base-driver/pull/230


lib/mjpeg.js
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Start reading the MJpeg stream and storing the last image
+   */
+  async start (serverTimeout = MJPEG_SERVER_TIMEOUT_MS) {
+    this.consumer = new MJpegConsumer();
+
+    // use the deferred pattern so we can wait for the start of the stream
+    // based on what comes in from an external pipe
+    let startPromise = new B((res, rej) => {
```
mykola-mokhnach: const

lib/mjpeg.js
```diff
@@ -0,0 +1,148 @@
...
+    this.consumer = null;
+    this.lastChunk = null;
+    this.updateCount = 0;
+  }
+
+  /**
+   * Start reading the MJpeg stream and storing the last image
+   */
+  async start (serverTimeout = MJPEG_SERVER_TIMEOUT_MS) {
+    this.consumer = new MJpegConsumer();
```
mykola-mokhnach: should we call `stop` here first ?

lib/mjpeg.js
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Start an mjpeg server for the purpose of testing, which just sends the same
+ * image over and over. Caller is responsible for closing the server.
+ * @param {int} port - port the server should listen on
+ * @param {int} [intMs] - how often the server should push an image
+ * @param {int} [times] - how many times the server should push an image before
+ * it closes the connection
+ * @returns {http.Server}
+ */
+function initMJpegServer (port, intMs = 300, times = 20) {
```
mykola-mokhnach: I'd rather put this helper method into tests

jlipps: it will be used by the drivers though, so it needs to be exported from appium-support

lib/mjpeg.js
jlipps: doesn't hurt to be safe

test/mjpeg-specs.js
```diff
@@ -0,0 +1,63 @@
...
+    stream.stop();
+    await B.delay(1000);
+    should.not.exist(stream.lastChunk);
+    stream.updateCount.should.eql(0);
+  });
+
+  it('should error out if the server does not send any images before a timeout', async function () {
+    let stream = new MJpegStream(MJPEG_SERVER_URL, _.noop);
+    await stream.start(0).should.eventually.be.rejectedWith(/never sent/);
+    stream.stop();
```
mykola-mokhnach: `stop` might not always be called if a test fails.

jlipps:
@mykola-mokhnach added png translation support

lib/mjpeg.js
```diff
@@ -0,0 +1,179 @@
...
+   * Get the base64-encoded version of the JPEG
+   * @returns {string}
+   */
+  get lastChunkBase64 () {
+    return _.isBuffer(this.lastChunk) ?
+      this.lastChunk.toString('base64') :
+      null;
+  }
+
+  async lastChunkPNG () {
```
mykola-mokhnach: a docstring  might be useful

jlipps:
@imurchie any thoughts or comments on this before i merge?

lib/mjpeg.js
```diff
@@ -0,0 +1,187 @@
...
+import http from 'http';
+import path from 'path';
+import B from 'bluebird';
+import Jimp from 'jimp';
+import MJpegConsumer from 'mjpeg-consumer';
+import mJpegServer from 'mjpeg-server';
+import { Writable } from 'stream';
+import { fs } from 'appium-support';
+
+const IMAGE_PATH = path.resolve(__dirname, '..', '..', 'test', 'images',
```
imurchie: `test/` is currently not included in the npm release, so this won't exist (fine for declaration, since `path.resolve` doesn't do anything to the contents, but not for the usage later).

jlipps: ah yes, good point. i'll find somewhere else for it to go.

mykola-mokhnach: you could encode the simplest one-pixel-like image into base64-string and put it directly there

jlipps:
ok @mykola-mokhnach @imurchie changes made, what do you think now?

PR #76, opened 2018-07-16
Move tests into e2e where they belong

imurchie:
These are not unit tests.


jlipps:
right you are, thanks

PR #77, opened 2018-07-19
add some jimp helper functionality

jlipps:
was finding that i was having to do lots of bluebird messiness just to use jimp in a few different places. this PR puts all that messiness in one place.

### DIFF
--- a/lib/image-util.js
+++ b/lib/image-util.js
@@ -1,8 +1,11 @@
 import _ from 'lodash';
+import Jimp from 'jimp';
 import { Buffer } from 'buffer';
 import { PNG } from 'pngjs';
 import B from 'bluebird';
 import { hasValue } from './util';
+
+const { MIME_JPEG, MIME_PNG, MIME_BMP } = Jimp;
 let cv = null;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/image-util.js
```diff
@@ -52,6 +55,37 @@ const AVAILABLE_MATCHING_FUNCTIONS = [
   'BruteForceSL2',
 ];
 
+/**
+ * Utility function to get a Jimp image object from buffer or base64 data. Jimp
+ * is a great library however it does IO in the constructor so it's not
+ * convenient for our async/await model.
+ *
+ * @param {Buffer|string} data - binary image buffer or base64-encoded image
+ * string
```
mykola-mokhnach: missing return description

lib/image-util.js
```diff
@@ -52,6 +55,37 @@ const AVAILABLE_MATCHING_FUNCTIONS = [
...
+ */
+async function getJimpImage (data) {
```
mykola-mokhnach: Perhaps `toJimpImage` ?

lib/image-util.js
jlipps: fixed

lib/image-util.js
jlipps: hmm, if this were a method on an object i would like that, but since it's just a bare function that feels confusing to me. open to other ideas though.

lib/image-util.js
```diff
@@ -52,6 +55,38 @@ const AVAILABLE_MATCHING_FUNCTIONS = [
...
+ * convenient for our async/await model.
+ *
+ * @param {Buffer|string} data - binary image buffer or base64-encoded image
+ * string
+ * @returns {Jimp} - the jimp image object
+ */
+async function getJimpImage (data) {
+  return await new B((resolve, reject) => {
+    if (!_.isString(data) && !_.isBuffer(data)) {
+      throw new Error("Must initialize jimp object with string or buffer");
```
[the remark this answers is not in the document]
jlipps: yes, good idea

jlipps:
published in `2.19.0`

PR #83, opened 2018-09-25
Do not check the list of supported versions

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11422

I would assume such verification should not be located in the library, but in the client code instead (appium-doctor in such case if it is necessary at all).

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -1,8 +1,7 @@
 import { exec } from 'teen_process';
 import os from 'os';
 
-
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@imurchie The test is fixed

DeChrish:
#11422 is fixed?

DeChrish:
https://github.com/appium/appium/issues/11422

PR #85, opened 2018-10-01
Update appium-gulp-plugins

imurchie:
This PR does two main things:
1. Updates `eslint-config-appium` and fixes any linting errors.
1. Updated `appium-gulp-plugins` and moves to Babel 7 and Gulp 4.

See https://github.com/appium/appium-gulp-plugins/pull/35 for more details.

### DIFF
--- a/test/zip-specs.js
+++ b/test/zip-specs.js
@@ -67,7 +67,7 @@ describe('#zip', function () {
 
     it('should stop iterating zipFile if onEntry callback returns false', async function () {
       let i = 0;
-      await zip.readEntries(zippedFilepath, async () => {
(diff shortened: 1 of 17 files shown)


test/zip-specs.js
```diff
@@ -67,7 +67,7 @@ describe('#zip', function () {
 
     it('should stop iterating zipFile if onEntry callback returns false', async function () {
       let i = 0;
-      await zip.readEntries(zippedFilepath, async () => {
+      await zip.readEntries(zippedFilepath, async () => { // eslint-disable-line require-await
```
mykola-mokhnach: is async needed here for the lambda?

PR #92, opened 2018-10-23
Properly handle folders while unzipping

mykola-mokhnach:
This fix should at least partially resolve https://travis-ci.org/appium/appium-support/jobs/444780357


mykola-mokhnach:
Ok, will change it to a functional. I am not sure why, but mockFS simply does not work properly with node10

PR #93, opened 2018-10-26
Unwrap w3c elements

dpgraham:
* `unwrapElement` should work for both [W3C elements](https://www.w3.org/TR/webdriver1/#elements) (`{"element-6066-11e4-a52e-4f735466cecf": "<ELEMENT-UUID>"}`) and JSONWP elements (`{"ELEMENT": "<ELEMENT-UUID>"}`)
* It prioritizes the W3C uuid when both are provided:
* e.g.)

```
{
  "ELEMENT": 1,
  "element-6066-11e4-a52e-4f735466cecf": 2 
}
```

will be unwrapped as 2.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -113,7 +113,7 @@ function safeJsonParse (obj) {
  *   { ELEMENT: 4 } becomes 4
  */
 function unwrapElement (el) {
-  if (_.isObject(el)) {
+  if (_.isPlainObject(el)) {
     if (el[W3C_WEB_ELEMENT_IDENTIFIER]) {
       return el[W3C_WEB_ELEMENT_IDENTIFIER];
     } else if (el.ELEMENT) {
--- a/lib/util.js
+++ b/lib/util.js
@@ -116,9 +116,9 @@ function unwrapElement (el) {
   if (_.isPlainObject(el)) {
     if (el[W3C_WEB_ELEMENT_IDENTIFIER]) {
       return el[W3C_WEB_ELEMENT_IDENTIFIER];
-    } else if (el.ELEMENT) {
-      return el.ELEMENT;
     }
+
+    return el.ELEMENT;
   }
   return el;
 }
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -111,8 +113,12 @@ function safeJsonParse (obj) {
  *   { ELEMENT: 4 } becomes 4
  */
 function unwrapElement (el) {
-  if (typeof el === 'object' && el.ELEMENT) {
-    return el.ELEMENT;
+  if (_.isObject(el)) {
```
mykola-mokhnach: isPlainObject

lib/util.js
```diff
@@ -111,8 +113,12 @@ function safeJsonParse (obj) {
...
+    if (el[W3C_WEB_ELEMENT_IDENTIFIER]) {
+      return el[W3C_WEB_ELEMENT_IDENTIFIER];
+    } else if (el.ELEMENT) {
```
mykola-mokhnach: else is not needed

dpgraham: What do you recommend we should do in the case that it's an object, but has neither of the keys ('ELEMENT', 'element-<w3c-udid>')? Should it be an exception?

lib/util.js
```diff
@@ -111,7 +113,11 @@ function safeJsonParse (obj) {
  *   { ELEMENT: 4 } becomes 4
  */
 function unwrapElement (el) {
-  if (typeof el === 'object' && el.ELEMENT) {
+  if (_.isPlainObject(el)) {
+    if (el[W3C_WEB_ELEMENT_IDENTIFIER]) {
```
mykola-mokhnach: I'd use a ternary operator here with `_.has` condition

mykola-mokhnach: actually, we can simply check

```
for (const propName of ['W3C_WEB_ELEMENT_IDENTIFIER', 'ELEMENT']) {
  if (_.has(el, propName) {
    return el[propName];
  }
}
return el;
```

PR #97, opened 2018-12-04
Add a helper to verify that two paths are pointing to the same destination (file or folder)

mykola-mokhnach:


mykola-mokhnach:
isSubPath does not use anything from fs and this one does. What is your proposal then?

imurchie:
My proposal would be that they both reside in the same place. It doesn't matter if they _use_ something internally, so much as does it make sense in use. If this function gets into `fs` then if we wanted to do something if it is the same file, and a different thing if it is a different file but within the same path, we would need functions from two different places to check what is essentially the same thing.

I guess conceptually I would propose adding a `path` module to `appium-support`, but it would also be fine to just make the two functions exist within the same namespace. If I had to choose a namespace I would say `util`, since the `isSameDestination` function uses `fs` functions but is not really a file system concern in the same way as the other functionality in that module.

mykola-mokhnach:
Ok, moved the method to util

PR #103, opened 2019-02-18
Fix the constant in hasAccess method

mykola-mokhnach:
The `fs` object here does not contain the necessary constants, so they all are equal to `undefined`. The constants should be imported from the system module instead or copied to our wrapper explicitly.

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -15,7 +15,7 @@ const md5 = B.promisify(md5file);
 let fs = {
   async hasAccess (path) {
     try {
-      await this.access(path, fs.F_OK | fs.R_OK);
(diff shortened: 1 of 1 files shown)


jlipps:
or use the property getter thing in es7

dpgraham:
Arg, I just noticed @jlipps comment. I merged this too early...

Is it okay to merge as is or should I revert and make the change?

mykola-mokhnach:
@dpgraham I assume that could be done in the other PR if needed. What do you think @imurchie ?

imurchie:
Yes. Another general PR would be a better place for that sort of thing anyway.

PR #105, opened 2019-03-01
Require from global if opencv4nodejs is not found locally

imurchie:
Node does not work as the code assumes it does. If a module is not found locally the system does not, by some magic, look at the globally installed modules (see https://github.com/nodejs/node/issues/3865).

I tried to find a published package that does this but didn't find anything I was particularly confident of. And the actual code is pretty simple.

This PR adds a little utility function (currently not exported, since we don't have any need for it anywhere else) to find the global package and load it.

This ought to finally solve https://github.com/appium/appium/issues/11865

### DIFF
--- a/lib/image-util.js
+++ b/lib/image-util.js
@@ -4,6 +4,10 @@ import { Buffer } from 'buffer';
 import { PNG } from 'pngjs';
 import B from 'bluebird';
 import { hasValue } from './util';
+import { exec } from 'teen_process';
+import path from 'path';
+import fs from './fs';
+
 
 const { MIME_JPEG, MIME_PNG, MIME_BMP } = Jimp;
 let cv = null;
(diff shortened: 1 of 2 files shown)


lib/image-util.js
```diff
@@ -87,18 +91,44 @@ async function getJimpImage (data) {
...
+ * @param {string} packageName - the name of the package to be required
+ */
+async function requirePackage (packageName) {
+  // see if we can get it locally
+  try {
+    return require(packageName);
+  } catch (ign) {}
+
+  // find the npm global root
+  const {stdout} = await exec('npm', ['root', '-g']);
```
mykola-mokhnach: this would probably fail on Windows, where npm is a batch script

mykola-mokhnach: https://github.com/appium/appium-doctor/pull/59/files

lib/image-util.js
```diff
@@ -87,18 +91,44 @@ async function getJimpImage (data) {
...
+  const globalNPMRoot = stdout.trim();
```
mykola-mokhnach: I'd log this output just in case...

lib/image-util.js
mykola-mokhnach: Also I'd prefer this exec call to be safe, e. g. if it returns non-zero code then we can at least get a proper error message

lib/image-util.js
```diff
@@ -87,18 +91,44 @@ async function getJimpImage (data) {
   });
 }
 
+/**
+ * Utility function to extend node functionality, allowing us to require
+ * modules that are installed globally
+ *
+ * @param {string} packageName - the name of the package to be required
```
mykola-mokhnach: missing return description

test/image-util-e2e-specs.js
```diff
@@ -66,7 +66,7 @@ describe('image-util', function () {
     });
 
     describe('getImagesMatches', function () {
-      it('should calculate the number of matches between two images', async function () {
+      it.only('should calculate the number of matches between two images', async function () { // eslint-disable-line
```
jlipps: is this `only` intentional?

imurchie: Yes. Trying to get Windows working on Travis.

imurchie:
Ok. Once I get it stable and correct, I will make it exported.

lib/image-util.js
```diff
@@ -87,18 +93,56 @@ async function getJimpImage (data) {
...
+  // see if we can get it locally
+  try {
+    log.debug(`Loading local package '${packageName}'`);
+    return require(packageName);
+  } catch (err) {
+    log.debug(`Unable to load local package '${packageName}': ${err.message}`);
+  }
+
+  // find the npm global root
+  const cmd = `npm${isWindows() ? '.cmd' : ''}`;
```
KazuCocoa: 👍

PR #107, opened 2019-03-16
Add utility method for versions comparison

mykola-mokhnach:
Such method might be useful in multiple modules


mykola-mokhnach:
Because I think such behaviour is safer in comparison to throwing an exception. if any of the version is not valid it would be logical to assume they are not equal (aka false)

imurchie:
But this doesn't only check equality. Is undefined < 1.2? Or is 1.2 > undefined?

I would prefer the caller to be responsible for making sure they have versions, or something like versions. And the compare function throw an error, or return a strict boolean.

mykola-mokhnach:
made it strict

PR #112, opened 2019-04-30
add tmpDir arg for tempdir to respect server tmpdir flag

KazuCocoa:
I've tested below patterns in [ruby_lib_core test suite](https://github.com/appium/ruby_lib_core/tree/master/test/functional) adding logs in `tempDir` module:

- Run a single device test
    - Android
        - UIA2
        - Espresso
    - iOS
        - XCUITest
- Run tests in parallel in single Appium server instance
    - Android x 3 emulators
        - UIA2
        - Espresso
    - iOS x 2 simulators
        - XCUITest
    - Android x 3, iOS x 2

The ruby_lib_core include screen recording, pull/push apks like https://github.com/appium/appium-android-driver/pull/538

Then, I ensured Appium log had no `/var/folders/xxxx`. They called `--tmp` arg.

----

Working for https://github.com/appium/appium/issues/12527

Appium can get `--tmp` as http://appium.io/docs/en/writing-running-appium/server-args/#server-flags. In some method, the flag is respected.

But methods calls `tempDir` module https://github.com/search?q=org%3Aappium+tempDir&type=Code does not respect it.

@jlipps suggested to me adding an arg to be able to refer the `--tmp` in `tempDir` modules.

After this, I should work to update `tempDir` module called in various repositories.

---

I'll update appium-android repository based on this PR in order to make sure if `--tmp` works expectedly with this change.

(So, currently this PR is just a draft)

----

related PRs:

- https://github.com/appium/appium-android-driver/pull/538
- https://github.com/appium/appium/pull/12585

### DIFF
--- a/lib/tempdir.js
+++ b/lib/tempdir.js
@@ -9,12 +9,12 @@ const RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
 
 /**
  * @param {?string} tempRootDirectory A path to root directory for temporary space.
- *                                    Defaults to process.env.APPIUM_TEMP_DIR.
- *                                    os.tmpdir() is called if process.env.APPIUM_TEMP_DIR is undefined.
+ *                                    Defaults to process.env.APPIUM_TMP_DIR.
+ *                                    os.tmpdir() is called if process.env.APPIUM_TMP_DIR is undefined.
  *
  * @returns A path to the available directory
  */
-async function tempDir (rootTmpDir = process.env.APPIUM_TEMP_DIR) {
+async function tempDir (rootTmpDir = process.env.APPIUM_TMP_DIR) {
   const now = new Date();
   const filePath = nodePath.join(rootTmpDir || os.tmpdir(),
     [
@@ -41,16 +41,14 @@ async function tempDir (rootTmpDir = process.env.APPIUM_TEMP_DIR) {
  * @param {string|Affixes} rawAffixes
  * @param {?string} defaultPrefix
  * @param {?string} tempRootDirectory Respect tempRootDirectory if this is provided.
- *                               Defaults to process.env.APPIUM_TEMP_DIR
- *                               {null|undefined} force ignore the rocess.env.APPIUM_TEMP_DIR.
+ *                               Defaults to process.env.APPIUM_TMP_DIR
+ *                               {null|undefined} force ignore the rocess.env.APPIUM_TMP_DIR.
  * @returns {string} A path to the temp directory in tempRootDirectory
  */
-async function path (rawAffixes, defaultPrefix, tempRootDirectory = process.env.APPIUM_TEMP_DIR) {
+async function path (rawAffixes, defaultPrefix, tempRootDirectory = process.env.APPIUM_TMP_DIR) {
   const affixes = parseAffixes(rawAffixes, defaultPrefix);
   const name = [affixes.prefix, affixes.suffix].join('');
   const tempDirectory = await tempDir(tempRootDirectory);
-  log.debug(`========[path] Use given tempRootDirectory in '${tempRootDirectory}'`);
-  log.debug(`========[path] Return: '${nodePath.join(tempDirectory, name)}'`);
   return nodePath.join(tempDirectory, name);
 }
 
@@ -64,13 +62,12 @@ async function path (rawAffixes, defaultPrefix, tempRootDirectory = process.env.
  *
  * @param {Affixes} affixes
  * @param {?string} tempRootDirectory Respect tempRootDirectory if this is provided.
- *                               Defaults to process.env.APPIUM_TEMP_DIR
- *                               {null|undefined} force ignore the rocess.env.APPIUM_TEMP_DIR.
+ *                               Defaults to process.env.APPIUM_TMP_DIR
+ *                               {null|undefined} force ignore the rocess.env.APPIUM_TMP_DIR.
  * @returns {OpenedAffixes}
  */
-async function open (affixes, tempRootDirectory = process.env.APPIUM_TEMP_DIR) {
+async function open (affixes, tempRootDirectory = process.env.APPIUM_TMP_DIR) {
   const filePath = await path(affixes, 'f-', tempRootDirectory);
-  log.debug(`========[open] Open: '${filePath}'`);
   try {
     let fd = await fs.open(filePath, RDWR_EXCL, 0o600);
     // opens the file in mode 384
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
I would rather use pass the global config directly to appium support if possible without adding additional parameters. Otherwise it will be necessary to modify all places in the client code where affected are called

jlipps:
@mykola-mokhnach are you thinking of something like a global var in appium-support that can be modified with a function like `setTmpDir`, which would then be called in Appium's main package with the value of config.tmpDir?

I think something like this could work, assuming that every module pulls from the same in-memory version of appium-support (in other words, that there is total deduplication of the appium-support package). not sure NPM can always guarantee this, though.

mykola-mokhnach:
@jlipps We could also set an environment variable to share this information between modules

lib/tempdir.js
```diff
@@ -39,7 +39,7 @@ async function tempDir () {
  * @param {?string} tempDirectory Respect tempDirectory if this is provided.
  * @returns {string} A path to the temp directory in tempDirectory
  */
-async function path (rawAffixes, defaultPrefix, tempDirectory = null) {
+async function path (rawAffixes, defaultPrefix, tempDirectory = process.env.APPIUM_TEMP_DIR) {
```
KazuCocoa: note for here:

https://github.com/appium/appium/issues/12527#issuecomment-488764185

jlipps:
yeah this makes sense to me

lib/tempdir.js
```diff
@@ -24,25 +27,61 @@ async function tempDir () {
...
+ * @property {string} suffix - suffix of the temp directory name
+ */
+
+/**
+ * @param {string|Affixes} rawAffixes
+ * @param {?string} defaultPrefix
+ * @param {?string} tempDirectory Respect tempDirectory if this is provided.
+ * @returns {string} A path to the temp directory in tempDirectory
+ */
+async function path (rawAffixes, defaultPrefix, respectTempDir = true) {
```
mykola-mokhnach: I think we don't need `respectTempDir` argument. If `process.env.APPIUM_TEMP_DIR` exists then we just use it by default

lib/tempdir.js
```diff
@@ -24,25 +27,61 @@ async function tempDir () {
...
+  let tempDirectory = respectTempDir ? process.env.APPIUM_TEMP_DIR : null;
+
   let affixes = parseAffixes(rawAffixes, defaultPrefix);
   let name = [affixes.prefix, affixes.suffix].join('');
-  let tempDirectory = await tempDir();
+  if (!tempDirectory) {
+    tempDirectory = await tempDir();
+    log.debug(`Create a new tempDir in '${tempDirectory}'`);
```
mykola-mokhnach: having logs here might make the server too chatty

lib/tempdir.js
```diff
@@ -24,25 +27,61 @@ async function tempDir () {
...
-async function open (affixes) {
-  let filePath = await path(affixes, 'f-');
+/**
+ * @typedef {Object} OpenedAffixes
+ * @property {string} path - The path to file
+ * @property {integer} fd - The file descriptor opened
+ */
+
+/**
+ *
```
mykola-mokhnach: A description string would be handy

lib/tempdir.js
KazuCocoa: Yes. I'll remove them after testing on my local to ensure all of `tempDir`

lib/tempdir.js
KazuCocoa: Thanks for your opinion :)
I also considered which way was better for us

lib/tempdir.js
```diff
@@ -24,25 +27,63 @@ async function tempDir () {
...
+ * @param {Affixes} affixes
+ * @param {?string} tempDirectory Respect tempDirectory if this is provided.
+ *                               Defaults to process.env.APPIUM_TEMP_DIR
+ *                               {null|undefined} force ignore the rocess.env.APPIUM_TEMP_DIR.
```
imurchie: process.

KazuCocoa:
I've tested as https://github.com/appium/appium-support/pull/112#issue-274805623 and created PRs which should fix for `--tmp` issue.

lib/tempdir.js
```diff
@@ -7,9 +7,21 @@ import log from './logger';
...
+ * - No `rootTmpDir`: `/var/folders/34/2222sh8n27d6rcp7jqlkw8km0000gn/T/xxxxxxxx.yyyy`
+ * - With `rootTmpDir = '/path/to/root'`: `/path/to/root/xxxxxxxx.yyyy`
+ *
+ * @param {?string} tempRootDirectory A path to a root directory for the temporary directory.
+ *                                    Defaults to process.env.APPIUM_TMP_DIR.
+ *                                    os.tmpdir() is called if process.env.APPIUM_TMP_DIR is undefined.
+ *
+ * @returns {string} A path to the temporary directory
+ */
+async function tempDir (rootTmpDir = process.env.APPIUM_TMP_DIR) {
```
mykola-mokhnach: I still don't understand why an additional argument is needed

KazuCocoa: yeah, here was still removing unused code

lib/tempdir.js
```diff
@@ -7,9 +7,21 @@ import log from './logger';
 
 const RDWR_EXCL = cnst.O_CREAT | cnst.O_TRUNC | cnst.O_RDWR | cnst.O_EXCL;
 
+/**
+ * Generate a temporary directory in os.tempdir() or tempRootDirectory.
+ * e.g.
+ * - No `process.env.APPIUM_TMP_DIR`: `/var/folders/34/2222sh8n27d6rcp7jqlkw8km0000gn/T/xxxxxxxx.yyyy`
+ * - With `process.env.APPIUM_TMP_DIR = '/path/to/root'`: `/path/to/root/xxxxxxxx.yyyy`
+ *
+ * @param {?string} tempRootDirectory A path to a root directory for the temporary directory.
```
mykola-mokhnach: this is not needed now

lib/tempdir.js
```diff
@@ -24,25 +36,59 @@ async function tempDir () {
...
+ * @param {string|Affixes} rawAffixes
+ * @param {?string} defaultPrefix
+ * @returns {string}  A path to the temporary directory with rawAffixes and defaultPrefix
+ */
 async function path (rawAffixes, defaultPrefix) {
-  let affixes = parseAffixes(rawAffixes, defaultPrefix);
-  let name = [affixes.prefix, affixes.suffix].join('');
-  let tempDirectory = await tempDir();
+  const affixes = parseAffixes(rawAffixes, defaultPrefix);
+  const name = [affixes.prefix, affixes.suffix].join('');
```
mykola-mokhnach: `${affixes.prefix}${affixes.suffix}`

lib/tempdir.js
```diff
@@ -48,7 +48,7 @@ async function tempDir () {
  */
 async function path (rawAffixes, defaultPrefix) {
   const affixes = parseAffixes(rawAffixes, defaultPrefix);
-  const name = `${affixes.prefix}${affixes.suffix}`;
```
mykola-mokhnach: how about 

```
${affixes.prefix || ''}${affixes.suffix || ''}
```

if we need to handle undefined?

KazuCocoa: ah, nice 👍

KazuCocoa:
Could you release the new version?

PR #118, opened 2019-06-24
Move the appium-ios-device plist functions here

umutuzgur:
`appium-ios-device` module needs certain plist operations and they are currently implemented there. The PR moves them here. The methods there will be removed after this PR is merged and a new version is published


.gitignore
```diff
@@ -5,3 +5,4 @@ coverage
 *.log
 .DS_Store
 package-lock.json*
+.vscode
```
jlipps: i tend to think users of specific editors should globally ignore their own editor configuration files, but don't feel too strongly about it.

(for example i use vim and make sure that my hidden vim files never get added to git repos regardless of what .gitignore says)

test/plist-specs.js
```diff
@@ -5,11 +5,12 @@ import { plist, tempDir, fs } from '../index.js';
 
 chai.should();
 
-const plistPath = path.resolve('test', 'assets', 'sample.plist');
+const binaryPlistPath = path.resolve('test', 'assets', 'sample.plist');
+const textPlistPath = path.resolve('test', 'assets', 'sample.plist');
```
jlipps: how can the binary and the text plist be the same file?

umutuzgur: It can't. I just forget to modify sometimes :D

.gitignore
umutuzgur: I don't fully agree. If you are used to using `git add .` instead of specific paths, it is nice to add these. I personally do that a lot and find these files in my commits just before I push them. I think it is common for people to use `git add .` if they don't use an IDE with good git integration

lib/plist.js
```diff
@@ -77,4 +85,47 @@ function parseBinaryPlist (data) {
   return bplistParse.parseBuffer(data);
 }
 
-export { parsePlistFile, updatePlistFile, createBinaryPlist, parseBinaryPlist };
+function getTextPlist (data) {
```
mykola-mokhnach: it would be nice to have docstrings for all these methods

lib/plist.js
```diff
@@ -77,4 +85,47 @@ function parseBinaryPlist (data) {
...
+  }
+  if (_.isBuffer(data) && PLIST_IDENTIFIER.BUFFER.compare(data, 0, PLIST_IDENTIFIER.BUFFER.length) === 0) {
+    return data.toString();
+  }
+  return null;
+}
+
+function getBinaryPlist (data) {
+  if (_.isString(data) && data.startsWith(BPLIST_IDENTIFIER.TEXT)) {
+    return Buffer.from(data);
```
mykola-mokhnach: is plist encoding always utf8?

lib/plist.js
```diff
@@ -77,4 +85,47 @@ function parseBinaryPlist (data) {
...
+
+function createPlist (object, binary = false) {
+  if (binary) {
+    return createBinaryPlist(object);
+  } else {
+    return xmlplist.build(object);
+  }
+}
+
+function parsePlist (data) {
```
mykola-mokhnach: isn't `parsePlistFile` doing the same?

lib/plist.js
umutuzgur: This case should nearly never happen and I assumed it would be `utf8` as it is the most common way

lib/plist.js
umutuzgur: This is not parsing a file but a string or a buffer

lib/plist.js
mykola-mokhnach: I can observe naming conflicts here. such plist are called xml plists instead of `text`

umutuzgur: I can take a look at that. I agree, it is nice to have documentation especially for the util methods

umutuzgur: Will rename that

.gitignore
jlipps: I always git add everything as well, I just have my vim stuff in my global gitignore so that it doesn't have to be in project-specific gitignores.

lib/plist.js
```diff
@@ -23,6 +23,12 @@ async function parseXmlPlistFile (plistFilename) {
   return xmlplist.parse(xmlContent);
 }
 
+/**
+ * Parses a file in xml or binary format of plist
+ * @param {string} plist The plist file path
+ * @param {boolean} mustExist If set to false, this method will return an empty object when the file doesn't exist
+ * @param {boolean} quiet If set to false, the plist path will be logged in debug level
```
mykola-mokhnach: missing return and throws descriptions

mykola-mokhnach: same below

lib/plist.js
```diff
@@ -58,6 +64,14 @@ async function parsePlistFile (plist, mustExist = true, quiet = true) {
   return obj;
 }
 
+/**
+ * Updates a plist file with the given fields
+ * @param {string} plist The plist file path
+ * @param {*} updatedFields The updated fields-value pairs
```
mykola-mokhnach: `{Object}`

lib/plist.js
```diff
@@ -68,13 +91,76 @@ async function updatePlistFile (plist, updatedFields, binary = true, mustExist =
...
+
+  let binaryPlist = getBinaryPlist(data);
+  if (binaryPlist) {
+    return parseBinaryPlist(binaryPlist)[0];
+  }
+
+  throw new Error(`Unknown type of plist, data: ${data.toString()}`);
+}
+
+export { parsePlistFile, parsePlist, createPlist, updatePlistFile, createBinaryPlist, parseBinaryPlist };
```
mykola-mokhnach: what if we follow the common style in naming (parsePlistFile <> parsePlist)?

umutuzgur: Do you have a suggestion? Lets collect the last comments so I can merge my PR

umutuzgur: @mykola-mokhnach Still waiting for you

PR #122, opened 2019-08-01
Add a helper method to compress file hierarchies based on glob pattern

mykola-mokhnach:
This method will be needed for adb, because Google has removed apk file management primitives from aapt2

### DIFF
--- a/lib/zip.js
+++ b/lib/zip.js
@@ -179,7 +179,8 @@ async function assertValidZip (filePath) {
 
 /**
  * @typedef {Object} ZipCompressionOptions
- * @property {number} level [9] - Compression level in range 0..9 (9 is the best)
+ * @property {number} level [9] - Compression level in range 0..9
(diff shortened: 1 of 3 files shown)


lib/zip.js
```diff
@@ -177,5 +177,51 @@ async function assertValidZip (filePath) {
   }
 }
 
-export { extractAllTo, readEntries, toInMemoryZip, _extractEntryTo, assertValidZip };
-export default { extractAllTo, readEntries, toInMemoryZip, assertValidZip };
+/**
+ * @typedef {Object} ZipCompressionOptions
+ * @property {number} level [9] - Compression level in range 0..9 (9 is the best)
```
imurchie: "best" is unclear. Does that mean most compression?

mykola-mokhnach: yep, I've changed the description

PR #123, opened 2019-08-01
Update travis build

imurchie:
Node 12 can handle `opencv4nodejs` and `mjpeg-consumer`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: xenial
+dist: trusty
 language: node_js
 node_js:
   - "10"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,24 +5,21 @@ node_js:
 addons:
   apt:
     sources:
-      - llvm-toolchain-trusty-5.0
+      - ubuntu-toolchain-r-test
     packages:
-      - clang-5.0
```
mykola-mokhnach: any reason to change to clang? Last time I tried to compile opencv with GCC it was failing doe to out of memory

mykola-mokhnach:
I actually set the dist to xenial: https://github.com/appium/appium-support/pull/122/files#diff-354f30a63fb0907d4ad57269548329e3R1

.travis.yml
imurchie: Mostly... because it works.

imurchie:
Ok. I'll let the dist change and then see if this is needed for building everything.

PR #137, opened 2019-10-26
feat: add function to stringify objects with Buffers

imurchie:
Add a general-purpose function to serialize objects while respecting Buffers. This is currently used in `appium-remote-debugger`, and it has a place in `appium-ios-device`

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -112,6 +112,40 @@ function safeJsonParse (obj) {
 }
 
 /*
+ * Stringifies the object passed in, converting Buffers into Strings for better
+ * display. This mimics JSON.stringify (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)
+ * except the `replacer` argument can only be a function.
+ *
+ * @param {object} obj - the object to be serialized
+ * @param {?function} replacer - function to transform the properties added to the
+ *                               serialized object
+ * @param {?number|string} space - used to insert white space into the output JSON
+ *                                 string for readability purposes. Defaults to 2
+ * returns {string} - the JSON object serialized as a string
+ */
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -111,6 +111,40 @@ function safeJsonParse (obj) {
...
+ * except the `replacer` argument can only be a function.
+ *
+ * @param {object} obj - the object to be serialized
+ * @param {?function} replacer - function to transform the properties added to the
+ *                               serialized object
+ * @param {?number|string} space - used to insert white space into the output JSON
+ *                                 string for readability purposes. Defaults to 2
+ * returns {string} - the JSON object serialized as a string
+ */
+function jsonStringify (obj, replacer, space = 2) {
```
mykola-mokhnach: replacer = null

lib/util.js
```diff
@@ -111,6 +111,40 @@ function safeJsonParse (obj) {
   }
 }
 
+/*
+ * Stringifies the object passed in, converting Buffers into Strings for better
+ * display. This mimics JSON.stringify (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)
+ * except the `replacer` argument can only be a function.
+ *
+ * @param {object} obj - the object to be serialized
+ * @param {?function} replacer - function to transform the properties added to the
```
mykola-mokhnach: it is important to mention which args the function accepts

mykola-mokhnach: and what is expected to return

imurchie: I think that is best left to the MDN entry mentioned in the description.

lib/util.js
imurchie: That's not the default, though. The default is defined below.

PR #138, opened 2019-10-30
Issue  #13515

anorimaki:
Pull request to fix https://github.com/appium/appium/issues/13515

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -85,27 +85,34 @@ let fs = {
       throw Error(`'${dir}' is not a valid root directory` + (errMsg ? `. Original error: ${errMsg}` : ''));
     }
 
+    let lastFileProcessed = B.resolve();
     return await new B((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
Thanks for the fix @anorimaki 
Could you please sign the CLA?

anorimaki:
> 
> 
> Thanks for the fix @anorimaki
> Could you please sign the CLA?

I have signed it but the check still is pending. I don't know what else I have to do.

mykola-mokhnach:
You should make sure the email address you used to sign the CLA is the same as the email address in the local config of git from which the PR has been prepared.
As an option you could fix the local email address and then rebase the PR

anorimaki:
> 
> 
> @anorimaki Seems like the bot still cannot figure it out :( . Would you mind opening a new PR with fixed email address? It must work then

https://github.com/appium/appium-support/pull/139
Let's see if there is more luck now :)

PR #145, opened 2019-12-09
build: do not cache npm dir

imurchie:
Our caching no longer works, and so the build fails. This takes a little longer, but at least runs things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,7 @@ addons:
     packages:
       - clang-5.0
 cache:
-  directories:
-    - "$(npm root -g)"
+  npm: false
 env:
   global:
     - CXX=clang++-5.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,17 +22,13 @@ install:
...
-        printf "while [ true ]; do\nsleep 30\necho 'Building OpenCV'\ndone" > ping.sh;
-        bash ping.sh &
-        echo $! > ping.pid;
-        npm install -g opencv4nodejs > /dev/null 2>&1;
-        kill `cat ping.pid`;
-      fi
+      printf "while [ true ]; do\nsleep 30\necho 'Building OpenCV'\ndone" > ping.sh;
+      bash ping.sh &
+      echo $! > ping.pid;
+      npm install -g opencv4nodejs > /dev/null 2>&1;
```
mykola-mokhnach: this will take forever %)

imurchie: Not quite forever, but longer than the cache. But slower is better than constant failure.

mykola-mokhnach: Perhaps we can preinstall opencv modules as debian packages, so then this operation would take much less time

imurchie: I've tried a couple of attempts to do this, to no avail. Any objection to merging this while I continue to try? I'd like to have coverage over this package, which is lacking right now.

mykola-mokhnach: I don't mind merging it, although Travis is slow and installing OpenCV stuff will make it even slower. I'd rather disable this by default

PR #146, opened 2019-12-10
build: fix node caching in travis

imurchie:
Re-enable global `node_modules` caching in Travis. Also remove the masking of the install process for `opencv4nodejs` as, in the event of an error in install/build, it makes the problem unclear.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,13 @@ node_js:
 addons:
   apt:
     sources:
-      - llvm-toolchain-r-test
+      - ubuntu-toolchain-r-test
     packages:
       - clang-5.0
 cache:
   npm: false
+  directories:
+    - "$(npm root -g)"
 env:
   global:
     - CXX=clang++-5.0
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -22,13 +24,17 @@ install:
...
   - if [[ `node --version` != v12* ]]; then
-      printf "while [ true ]; do\nsleep 30\necho 'Building OpenCV'\ndone" > ping.sh;
-      bash ping.sh &
-      echo $! > ping.pid;
-      npm install -g opencv4nodejs > /dev/null 2>&1;
-      kill `cat ping.pid`;
+      if [[ $(npm ls --depth 1 -g opencv4nodejs) =~ "── opencv4nodejs@" ]]; then
+        echo 'Already installed opencv4nodejs';
+      else
+        npm install -g opencv4nodejs;
```
mykola-mokhnach: travis was killing the job previously because the build command was printing too much of data into the log and/or was printing no data at all within 30 min making Travis think it's frozen. That is why the trick with ping.sh was needed

imurchie: It seems fine now. I've run it a number of times without running over the log limit.

PR #148, opened 2019-12-16
feat: Add synchronous rimraf

mykola-mokhnach:


lib/fs.js
```diff
@@ -25,6 +25,7 @@ let fs = {
   },
   exists (path) { return this.hasAccess(path); },
   rimraf: B.promisify(rimraf),
+  rimrafSync (path, opts) { return rimraf.sync(path, opts); },
```
imurchie: I don't have a preference, but this could probably also be
```js
rimrafSync: rimraf.sync.bind(rimraf),
```

mykola-mokhnach: yep, I like this approach better

PR #149, opened 2019-12-17
feat: add timing API

imurchie:
We have a number of different mechanisms for getting the duration of operations. This is a first pass at making an extensible API for our use. Once the Performance API is out of super-experimental in Node, this could encapsulate that too, without needing application code to be messed with.

### DIFF
--- a/lib/timing.js
+++ b/lib/timing.js
@@ -1,0 +1,85 @@
+import _ from 'lodash';
+
+
+const DURATION_SECONDS = 's';
+const DURATION_MILLIS = 'ms';
+const DURATION_NANOS = 'ns';
+
+const NS_PER_SEC = 1e9;
+const NS_PER_MS = 1e6;
+
+const CONVERSION_FACTORS = {
+  [DURATION_SECONDS]: NS_PER_SEC,
+  [DURATION_MILLIS]: NS_PER_MS,
+  [DURATION_NANOS]: 1,
+};
+
+/**
+ * Get the start time, to be used when getting the duration
+ *
+ * @return {number|Array<number>} a representation of the start time
+ */
+function getStartTime () {
+  // once Node 10 is no longer supported, this check can be removed
+  if (_.isFunction(process.hrtime.bigint)) {
+    return process.hrtime.bigint();
+  }
+  return process.hrtime();
+}
+
+/**
+ * @typedef {Object} GetDurationOptions
+ *
+ * @property {?string} units - the units to convert to. Can be one of
+ *     - 's' - seconds (default)
+ *     - 'ms' - milliseconds
+ *     - 'ns' - nanoseconds
+ * @property {?boolean} round - whether or not the result should be rounded down
+ */
+
+/**
+ * Get the duration based on a start time
+ *
+ * @param {number|Arrary<number>} startTime - the value returned from calling
+ *     getStartTime`
+ * @param {?string|GetDurationOptions} opts - options for conversion of duration.
+ *     Can be a String representing the units ('s', 'ms', or 'ns') or an Object
+ * @return {number} a number representing the duration, in the specified unit
+ */
+function getDuration (startTime, opts = {}) {
+  if (_.isString(opts)) {
+    opts = {
+      units: opts,
+    };
+  }
+  const {
+    units = DURATION_SECONDS,
+    round = true,
+  } = opts;
+
+  if (!_.keys(CONVERSION_FACTORS).includes(units)) {
+    throw new Error(`Unknown unit for duration conversion: '${units}'`);
+  }
+
+  let nanoDuration;
+  if (_.isArray(startTime)) {
+    // startTime was created using process.hrtime()
+    const [seconds, nanos] = process.hrtime(startTime);
+    nanoDuration = seconds * NS_PER_SEC + nanos;
+  } else if (typeof startTime === 'bigint' && _.isFunction(process.hrtime.bigint)) {
+    // startTime was created using process.hrtime.bigint()
+    const endTime = process.hrtime.bigint();
+    // get the difference, and convert to number
+    nanoDuration = Number(endTime - startTime);
+  } else {
+    throw new Error(`Unable to get duration. Start time '${startTime}' cannot be handled`);
+  }
+
+  const duration = nanoDuration / CONVERSION_FACTORS[units];
+  return round ? Math.round(duration) : duration;
+}
+
+
+export {
+  getStartTime, getDuration, DURATION_MILLIS, DURATION_SECONDS, DURATION_NANOS,
+};
--- a/lib/timing.js
+++ b/lib/timing.js
@@ -1,63 +1,65 @@
 import _ from 'lodash';
 
 
-const DURATION_S = 's';
-const DURATION_MS = 'ms';
-const DURATION_NS = 'ns';
-
-const NS_PER_SEC = 1e9;
+const NS_PER_S = 1e9;
 const NS_PER_MS = 1e6;
-
-const CONVERSION_FACTORS = {
-  [DURATION_S]: NS_PER_SEC,
-  [DURATION_MS]: NS_PER_MS,
-  [DURATION_NS]: 1,
-};
 
 
 /**
  * Class representing a duration, encapsulating the number and units.
  */
 class Duration {
-  constructor (duration, units) {
+  constructor (duration) {
     this._duration = duration;
-    this._units = units;
   }
 
   get duration () {
     return this._duration;
   }
 
-  get units () {
-    return this._units;
+  /**
+   * Get the duration as nanoseconds
+   *
+   * @returns {number} The duration as nanoseconds
+   */
+  get asNanoSeconds () {
+    return this.duration;
+  }
+
+  /**
+   * Get the duration converted into milliseconds
+   *
+   * @returns {number} The duration as milliseconds
+   */
+  get asMilliSeconds () {
+    return this.duration / NS_PER_MS;
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Start the timer
+   *
+   * @return {number|Array<number>} a representation of the start time
+   */
+  start () {
+    // once Node 10 is no longer supported, this check can be removed
+    this.startTime = _.isFunction(process.hrtime.bigint)
```
jlipps: `startTime` could be memoized during module load so it doesn't need to happen every time this class is used. Probably a very small optimization, but presumably when timing things we want to have as little overhead as possible.

jlipps: and by `startTime`, i mean, the function used to get the start time

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * @typedef {Object} GetDurationOptions
+   *
+   * @property {?string} units - the units to convert to. Can be one of
+   *     - 's' - seconds (default)
+   *     - 'ms' - milliseconds
+   *     - 'ns' - nanoseconds
+   * @property {?boolean} round - whether or not the result should be rounded down
```
mykola-mokhnach: I'd rather have a number of how many decimal digits after the dot we have to keep in the rounded number

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+   */
+
+  /**
+   * Get the duration since the timer was started
+   *
+   * @param {?string|GetDurationOptions} opts - options for conversion of duration.
+   *     Can be a String representing the units ('s', 'ms', or 'ns') or an Object
+   * @return {number} a number representing the duration, in the specified unit
```
mykola-mokhnach: how about having a separate object for the Duration itself, which can be represented in the readable string format?

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+        units: opts,
+      };
+    }
+    const {
+      units = DURATION_SECONDS,
+      round = true,
+    } = opts;
+
+    if (!_.keys(CONVERSION_FACTORS).includes(units)) {
+      throw new Error(`Unknown unit for duration conversion: '${units}'`);
```
mykola-mokhnach: How about explicitly enumerating the supported units in the error message?

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+    let nanoDuration;
+    if (_.isArray(this.startTime)) {
+      // startTime was created using process.hrtime()
+      const [seconds, nanos] = process.hrtime(this.startTime);
+      nanoDuration = seconds * NS_PER_SEC + nanos;
+    } else if (typeof this.startTime === 'bigint' && _.isFunction(process.hrtime.bigint)) {
+      // startTime was created using process.hrtime.bigint()
+      const endTime = process.hrtime.bigint();
+      // get the difference, and convert to number
+      nanoDuration = Number(endTime - this.startTime);
```
mykola-mokhnach: what's the point to use bigint if we anyway need to cast to Number?

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+  constructor (autoStart = false) {
+    this.startTime = autoStart ? this.start() : null;
+  }
+
+  /**
+   * Start the timer
+   *
+   * @return {number|Array<number>} a representation of the start time
+   */
+  start () {
```
mykola-mokhnach: I'd rather make this method to return the self instance and create a separate getter for `startTime`

mykola-mokhnach: then the constructor would not need to have autoStart argument, since we could write something like const timer = new Timer.start()

lib/timing.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+
+class Timer {
+  constructor (autoStart = false) {
+    this.startTime = autoStart ? this.start() : null;
+  }
+
+  /**
+   * Start the timer
```
mykola-mokhnach: do we care if this function would be called more than once?

test/timing-specs.js
```diff
@@ -0,0 +1,210 @@
+import _ from 'lodash';
```
mykola-mokhnach: 👍 
I wish I could write the tests so nicely

lib/timing.js
imurchie: The high resolution timestamp needs to be handled in a bigint, but the amount of time represented by a duration is smaller, and can be handled by a normal number.

It's the way Node's API is moving (https://nodejs.org/api/process.html#process_process_hrtime_time):
```
This is the legacy version of process.hrtime.bigint() before bigint was introduced in JavaScript.
```

lib/timing.js
imurchie: That would be possible now, too, since the constructor returns an instance.

lib/timing.js
imurchie: I'm not sure what you mean. There will be no decimal digits.

lib/timing.js
imurchie: I guess it would not get back the instance. Thanks.

lib/timing.js
mykola-mokhnach: I mean instead of having boolean flag to round let say `3.14152` seconds to `3` we could have a number of, for example, `2`, which is going to return `3.14` as the result or `0` for just `3`

imurchie: I would say that if you wanted `3.14s`, what you really want is `3140ms`.

imurchie:
Added the ability to just print the `Timer` itself, without calling `getDuration()`. E.g.,
```js
const timer = new timing.Timer().start();
// ...
log.debug(`The operation took ${timer}`);
```
outputs
```
The operation took 42ms
```

lib/timing.js
```diff
@@ -0,0 +1,144 @@
...
+      nanoDuration = Number(endTime - this.startTime);
+    } else {
+      throw new Error(`Unable to get duration. Start time '${this.startTime}' cannot be parsed`);
+    }
+
+    const duration = nanoDuration / CONVERSION_FACTORS[units];
+    return new Duration(round ? Math.round(duration) : duration, units);
+  }
+
+  toString () {
```
mykola-mokhnach: is it fine to throw an exception here if `start` was not called before?

imurchie: I think so. It is a programming error if it does. But I guess some odd default string would indicate that, too, if less dramatically.

lib/timing.js
```diff
@@ -0,0 +1,144 @@
...
+    return `${this.duration}${this.units}`;
+  }
+}
+
+class Timer {
+  /**
+   * Creates a timer
+   *
+   * @param {?string} units - the default units for conversion
```
mykola-mokhnach: I still cannot understand why units are needed. We could simply keep the absolute number of seconds/nanoseconds and have asSeconds, asNanos, etc. methods like TimeSpan in C# or Duration in Java have

imurchie:
Alright. Not sure why it's preferable, but moved the conversions in the `Duration` class, so now it is used something like
```js
const timer = new Timer().start();

// ...

log.debug(`This took ${timer.getDuration().asSeconds.toFixed(3)}s`)
```

lib/timing.js
```diff
@@ -0,0 +1,119 @@
...
+
+const NS_PER_S = 1e9;
+const NS_PER_MS = 1e6;
+
+
+/**
+ * Class representing a duration, encapsulating the number and units.
+ */
+class Duration {
+  constructor (duration) {
```
mykola-mokhnach: I's rather rename the property to `nanos`

lib/timing.js
```diff
@@ -0,0 +1,119 @@
...
+    this._duration = duration;
+  }
+
+  get duration () {
```
mykola-mokhnach: isn't this getter redundant now?

imurchie:
Basically. But mirroring the internal, read-only variable.

lib/timing.js
```diff
@@ -0,0 +1,119 @@
...
+    this._startTime = _.isFunction(process.hrtime.bigint)
+      ? process.hrtime.bigint()
+      : process.hrtime();
+    return this;
+  }
+
+  /**
+   * Get the duration since the timer was started
+   *
+   * @return {Duration} the duration, in the specified units
```
mykola-mokhnach: the param description should be updated

lib/timing.js
```diff
@@ -0,0 +1,119 @@
...
+  getDuration () {
+    if (_.isNull(this.startTime)) {
+      throw new Error(`Unable to get duration. Timer was not started`);
+    }
+
+    let nanoDuration;
+    if (_.isArray(this.startTime)) {
+      // startTime was created using process.hrtime()
+      const [seconds, nanos] = process.hrtime(this.startTime);
+      nanoDuration = seconds * NS_PER_S + nanos;
```
mykola-mokhnach: we can return immediately here. There is no need for the `nanoDuration` intermediate variable

lib/timing.js
```diff
@@ -0,0 +1,119 @@
...
+
+
+const NS_PER_S = 1e9;
+const NS_PER_MS = 1e6;
+
+
+/**
+ * Class representing a duration, encapsulating the number and units.
+ */
+class Duration {
```
mykola-mokhnach: In the future we could make the Duration class more advanced (e.g. add +/- operations, factory methods to create duration objects from different units (aka `Duration.fromSeconds(x)`, etc)) and use it in other scenarios. 
Or use the one, from [moment.js ](https://momentjs.com/docs/#/durations/) lib.

Although, the current implementation is pretty enough for the task it is supposed to solve

PR #158, opened 2020-02-01
chore: add todo to get rid of mkdirp dep

imurchie:
Node 10.12 introduced the `recursive` option to `fs.mkdir`(https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback).


jimmywarting:
remove it now?

PR #159, opened 2020-02-06
feat: export pluralize functionality

imurchie:
Add export of https://github.com/blakeembrey/pluralize through the `util` module.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -5,6 +5,7 @@ import path from 'path';
 import fs from './fs';
 import semver from 'semver';
 import { quote as shellQuote } from 'shell-quote';
+import pluralizeLib from 'pluralize';
 
 
 const W3C_WEB_ELEMENT_IDENTIFIER = 'element-6066-11e4-a52e-4f735466cecf';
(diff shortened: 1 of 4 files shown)


lib/util.js
```diff
@@ -338,10 +339,21 @@ function unleakString (s) {
   return ` ${s}`.substr(1);
 }
 
+/**
+ * Get the form of a word appropriate to the count
+ *
+ * @param {string} word - The word to pluralize
+ * @param {number} count - How many of the word exist
+ * @param {?boolean} inclusive - Whether to prefix with the number (e.g., 3 ducks)
```
mykola-mokhnach: `inclusive [false]`
I'd also think about putting all options into a dictionary, so we could extend it later if needed

imurchie: Sure.

lib/util.js
```diff
@@ -338,10 +339,33 @@ function unleakString (s) {
...
+ * @property {?boolean} inclusive [false] - Whether to prefix with the number (e.g., 3 ducks)
+ */
+
+/**
+ * Get the form of a word appropriate to the count
+ *
+ * @param {PluralizeOptions} options - options for word pluralization
+ * @returns {string} The word pluralized according to the number
+ */
+function pluralize (options = {}) {
```
mykola-mokhnach: the word and the count may stay as ordinary args, since they are always required

lib/util.js
```diff
@@ -342,23 +342,27 @@ function unleakString (s) {
...
  * @returns {string} The word pluralized according to the number
  */
-function pluralize (options = {}) {
-  const {
-    word,
-    count,
-    inclusive = false,
-  } = options;
+function pluralize (word, count, options = {}) {
+  let inclusive = false;
```
mykola-mokhnach: 👍

PR #163, opened 2020-03-11
feat: Perform fine-tuning of toInMemoryZip helper

mykola-mokhnach:
Streamed base64 encoding will help to save memory, since sometimes the amount of data passed to `toInMemoryZip` helper is super large. Also having some precautions like maxSize wouldn't hurt us in any case

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -389,6 +389,10 @@ function pluralize (word, count, options = {}) {
  * or the source file is too big to be read completely into the memory
  */
 async function toInMemoryBase64 (srcPath, opts = {}) {
+  if (!(await fs.exists(srcPath)) || (await fs.stat(srcPath)).isDirectory()) {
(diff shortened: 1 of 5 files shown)


lib/util.js
```diff
@@ -6,9 +6,11 @@ import fs from './fs';
 import semver from 'semver';
 import { quote as shellQuote } from 'shell-quote';
 import pluralizeLib from 'pluralize';
-
+import stream from 'stream';
+import { Base64Encode } from 'base64-stream';
 
 const W3C_WEB_ELEMENT_IDENTIFIER = 'element-6066-11e4-a52e-4f735466cecf';
+const GiB = 1024 * 1024 * 1024;
```
KazuCocoa: 👍

mykola-mokhnach:
It looks like there is a problem with streamed base64 encoding. Reported the issue: https://github.com/mazira/base64-stream/issues/12

PR #168, opened 2020-03-18
feat: Add file name sanitizer to the fs module

mykola-mokhnach:


mykola-mokhnach:
> Will Appium usage always include a replacement option? 

I wanted to make it configurable. By default this is just an empty string (means invalid chars are just cut off: https://github.com/parshap/node-sanitize-filename#readme)

imurchie:
Yeah, saw that in the docs. But if, say, Appium generally uses `-` as the replacement, it could be defaulted here while still maintaining the configurability.

PR #169, opened 2020-03-21
feat: returns similarity  score by getImageOccurrence

KazuCocoa:
https://github.com/appium/appium/issues/13883

Returns similarity score as the result of `getImageOccurrence`


lib/image-util.js
```diff
@@ -417,6 +417,8 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
  * @property {?Buffer} visualization The visualization of the matching result
  * represented as PNG image buffer. On this image the matching
  * region is highlighted with a rectangle.
+ * @property {float} score The similarity score as a float number in range [0.0, 1.0].
```
mykola-mokhnach: it is a number. javascript is kind'a weird, so `1.0` there equals to `1`

PR #179, opened 2020-04-22
feat: add getLockFileGuard method to util

jlipps:
the idea is to support guarding behaviors across multiple processes: https://github.com/appium/appium/pull/13463#pullrequestreview-304978624


mykola-mokhnach:
I would rather use some ready solution like https://www.npmjs.com/package/lockfile and create a wrapper over it inside appium-support

lib/fs.js
```diff
@@ -158,6 +159,8 @@ for (const s of simples) {
 const syncFunctions = [
   'createReadStream',
   'createWriteStream',
+  'writeFileSync',
```
mykola-mokhnach, line 162: so we still need these?

lib/util.js
```diff
@@ -443,11 +444,49 @@ async function toInMemoryBase64 (srcPath, opts = {}) {
...
+ * Create an async function which, when called, will not proceed until a certain file is no
+ * longer present on the system. This allows for preventing concurrent behavior across processes
+ * using a known lockfile path.
+ *
+ * @param {string} lockFile The full path to the file used for the lock
+ * @param {number} timeout [120] The max time in seconds to wait for the lock
+ * @returns {AsyncFunction} async function that takes another async function defining the locked
+ * behavior
+ */
+function getLockFileGuard (lockFilePath, timeout = 120) {
```
mykola-mokhnach, line 457: it makes sense to make options a dictionary, so we can add more of them later

lib/util.js
```diff
@@ -443,11 +444,49 @@ async function toInMemoryBase64 (srcPath, opts = {}) {
...
+  const lock = B.promisify(lockfile.lock);
+  const checkSync = B.promisify(lockfile.checkSync);
+  const unlock = B.promisify(lockfile.unlock);
+
+  return async (behavior) => {
+    try {
+      // if the lockfile doesn't exist, lock it synchronously to make sure no other call
+      // on the same spin of the event loop can also initiate a lock. If the lockfile does exist
+      // then just use the regular async 'lock' method which will wait on the lock.
+      if (!checkSync(lockFilePath)) {
```
mykola-mokhnach, line 468: this call is async

lib/util.js
```diff
@@ -443,11 +444,49 @@ async function toInMemoryBase64 (srcPath, opts = {}) {
...
+ * longer present on the system. This allows for preventing concurrent behavior across processes
+ * using a known lockfile path.
+ *
+ * @param {string} lockFile The full path to the file used for the lock
+ * @param {number} timeout [120] The max time in seconds to wait for the lock
+ * @returns {AsyncFunction} async function that takes another async function defining the locked
+ * behavior
+ */
+function getLockFileGuard (lockFilePath, timeout = 120) {
+  const lockSync = B.promisify(lockfile.lockSync);
```
mykola-mokhnach, line 458: something is not right here. How lockSync can be promisified?

jlipps:
comments addressed!

lib/util.js
```diff
@@ -443,11 +444,51 @@ async function toInMemoryBase64 (srcPath, opts = {}) {
...
+/**
+ * Create an async function which, when called, will not proceed until a certain file is no
+ * longer present on the system. This allows for preventing concurrent behavior across processes
+ * using a known lockfile path.
+ *
+ * @param {LockFileOptions} opts
+ * @returns {AsyncFunction} async function that takes another async function defining the locked
+ * behavior
+ */
+function getLockFileGuard ({lockFile, timeout = 120}) {
```
mykola-mokhnach, line 461: the first parameter is always required. I meant something like

```
async function getLockFileGuard (lockFile, opts = {})
```

jlipps: oh i see, right.

test/util-e2e-specs.js
```diff
@@ -33,4 +34,103 @@ describe('#util', function () {
     });
   });
 
+  describe('getLockFileGuard()', function () {
+    let tmpRoot;
+    let lockFile;
+    let testFile;
+
+    async function lockExists () {
```
mykola-mokhnach, line 42: should we use appropriate methods from `lockfile` module instead or have wrappers over them to check if lock exists?

jlipps: yeah, good idea.

lib/fs.js
```diff
@@ -150,6 +150,7 @@ let fs = {
 const simples = [
   'open', 'close', 'access', 'readFile', 'writeFile', 'write', 'read',
   'readlink', 'chmod', 'unlink', 'readdir', 'stat', 'rename', 'lstat',
+  'appendFile'
```
mykola-mokhnach, line 153: do we need this?

test/util-e2e-specs.js
```diff
@@ -33,4 +35,95 @@ describe('#util', function () {
     });
   });
 
+  describe('getLockFileGuard()', function () {
+    let tmpRoot;
+    let lockFile;
+    let testFile;
+    const lockExists = () => B.promisify(lockfile.check)(lockFile);
```
mykola-mokhnach, line 42: does it make sense to make this check public so we could check the lock in other modules without importing lockfile explicitly?

jlipps: yeah, good idea. done.

PR #180, opened 2020-04-26
feat: Make it possible to replace sensitive info in logs

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/7622

The idea is to provide a command line parameter to the server, which contains a path to a JSON file having one or more secure values replacement rules. The example of such config:

```
[
     "simple text to replace",
      { "pattern": "very.+secure", "flags": "s" },
      { "text": "caseInsensitiveText", "flags": "i", "replacer": "__secure__" }
]
```

Once loaded statically the set of parsed rules will be applied to each logged string and will replace all occurrences of the given text/pattern with either the default or the given `replacer` value to hide sensitive info from log readers.


test/log-internals-specs.js
```diff
@@ -0,0 +1,86 @@
...
+    const replacer2 = '***';
+    const issues = await preprocessor.loadRules([
+      null,
+      { flags: 'i' },
+      { pattern: '^:(', replacer: replacer2 },
+    ]);
+    issues.length.should.eql(3);
+    preprocessor.rules.length.should.eql(0);
+    preprocessor.preprocess(':yolo" yo Yolo yyolo').should.eql(':yolo" yo Yolo yyolo');
+  });
```
KazuCocoa, line 76: Adding one more `{ pattern: '^:', text: 'yolo'. replacer: replacer2 },` case to show the text/pattern's priority?

mykola-mokhnach: test added

mykola-mokhnach:
cc @imurchie

PR #196, opened 2020-07-10
chore: Validate Content-Length response header for file downloads

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/14534


test/net-e2e-specs.js
```diff
@@ -0,0 +1,29 @@
+import chai from 'chai';
```
KazuCocoa, line 1: 👍

PR #198, opened 2020-07-14
fix: Specify node version > 10.0.0 

Wangszzju:
The [stream pipeline api](https://nodejs.org/docs/latest-v8.x/api/stream.html) is not supported in node version 8.x.

We should tell the user it's not supported

Otherwise, when we use the bluebird promisify api, we get the error

```
expecting a function but got [object Undefined]
```


mykola-mokhnach:
Thanks @Wangszzju. Could you please sign the CLA?

PR #206, opened 2020-09-02
feat: Add an option to recover a "broken" file lock

mykola-mokhnach:
There could be cases where a process has been forcefully terminated without a chance to clean up pending locks: https://github.com/npm/lockfile/issues/26


KazuCocoa (approved):
seems reasonable to me

PR #210, opened 2020-12-18
test: unquarantine opencv tests since they work on node 12

jlipps:


.travis.yml
```diff
@@ -20,16 +20,16 @@ env:
     - CXX=clang++-5.0
     - CC=clang-5.0
     - _FORCE_LOGS=1
+    - NODE_PATH="$(npm root -g)"
 install:
-  # on node 12 opencv4nodejs and mjpeg-consumer cannot be installed
-  # also handle possible travis caching
+  # handle possible travis caching
+  - if [[ $(npm ls --depth 1 -g opencv4nodejs) =~ "── opencv4nodejs@" ]]; then
```
mykola-mokhnach, line 26: These tests will take forever on Travis. I would rather think about moving them to Azure like we already did for other reps

jlipps: yeah that's a better idea

PR #212, opened 2021-01-05
feat: implement multiple image detection

khaong:
Hi,

For my appium use case, there may be multiple places on the screen which has the same image and I'd like to find them all.

I am not a javascript dev by trade, so would definitely welcome any comments on this!

Thanks

Alex


lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+    // Below error message, `Cannot find any occurrences` is referenced in find by image
+    throw new Error(`Cannot find any occurrences of the partial image in the full image. ` +
+      `Original error: ${e}`);
+  }
+
+  if (visualize) {
+    results.forEach((result) => {
+      highlightRegion(fullImg, result.rect);
+    });
+    visualization = await cv.imencodeAsync('.png', fullImg);
```
khaong, line 578: Q: should we be returning a visualisation per OccurenceResult?

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+ * @property {?Buffer} visualization The visualization of the matching result
+ * represented as PNG image buffer. On this image the matching
+ * region is highlighted with a rectangle.
+ * @property {number} score The similarity score as a float number in range [0.0, 1.0].
+ * 1.0 is the highest score (means both images are totally equal).
+ */
+
+/**
+ * @typedef {Object} OccurrenceResults
+ * @property {OccurrenceResult[]} results The array of matching OccurenceResults
```
mykola-mokhnach, line 493: `Array<OccurrenceResult>`

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+    if (results.length === 0) {
+      // Below error message, `Cannot find any occurrences` is referenced in find by image
+      throw new Error(`Cannot find any occurrences of the partial image in the full ` +
+                      `image above the threshold of ${threshold}. Highest match value ` +
+                      `found was ${minMax.maxVal}`);
+    }
+  } catch (e) {
+    // Below error message, `Cannot find any occurrences` is referenced in find by image
+    throw new Error(`Cannot find any occurrences of the partial image in the full image. ` +
+      `Original error: ${e}`);
```
mykola-mokhnach, line 571: e.message

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+ */
+async function getImageOccurrences (fullImgData, partialImgData, options = {}) {
+  await initOpenCV();
+
+  const { visualize = false, threshold = DEFAULT_MATCH_THRESHOLD } = options;
+  const [fullImg, partialImg] = await B.all([
+    cv.imdecodeAsync(fullImgData),
+    cv.imdecodeAsync(partialImgData)
+  ]);
+  let results = [];
```
mykola-mokhnach, line 518: I prefer const where possible

mykola-mokhnach: same below

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+  let visualization = null;
+
+  try {
+    const matched = await fullImg.matchTemplateAsync(partialImg, cv.TM_CCOEFF_NORMED);
+    const minMax = await matched.minMaxLocAsync();
+
+    let nonZero = matched.threshold(threshold, 1, cv.THRESH_BINARY).convertTo(cv.CV_8U).findNonZero();
+    let arr = matched.getDataAsArray();
```
mykola-mokhnach, line 526: can we have better var name?

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+
+    let matches = [];
+
+    nonZero.forEach((element) => {
+      let hasNoNeighbours = true;
+
+      matches.some((match) => {
+        let distanceBetweenPoints = distance(match, element);
+        if (distanceBetweenPoints <= 10) {
```
mykola-mokhnach, line 535: what is this magic number?

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+  try {
+    const matched = await fullImg.matchTemplateAsync(partialImg, cv.TM_CCOEFF_NORMED);
+    const minMax = await matched.minMaxLocAsync();
+
+    let nonZero = matched.threshold(threshold, 1, cv.THRESH_BINARY).convertTo(cv.CV_8U).findNonZero();
+    let arr = matched.getDataAsArray();
+
+    let matches = [];
+
+    nonZero.forEach((element) => {
```
mykola-mokhnach, line 530: `for .. of` loop is more preferable

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+        width: partialImg.cols,
+        height: partialImg.rows
+      };
+
+      results.push(result);
+    });
+
+    if (results.length === 0) {
+      // Below error message, `Cannot find any occurrences` is referenced in find by image
+      throw new Error(`Cannot find any occurrences of the partial image in the full ` +
```
mykola-mokhnach, line 564: Why this throw must be inside of try/catch block?

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+      highlightRegion(fullImg, result.rect);
+    });
+    visualization = await cv.imencodeAsync('.png', fullImg);
+  }
+
+  return { results, visualization };
+}
+
+function distance (point1, point2) {
+  let a2 = Math.pow((point1.x - point2.x), 2);
```
mykola-mokhnach, line 585: const

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+    results.forEach((result) => {
+      highlightRegion(fullImg, result.rect);
+    });
+    visualization = await cv.imencodeAsync('.png', fullImg);
+  }
+
+  return { results, visualization };
+}
+
+function distance (point1, point2) {
```
mykola-mokhnach, line 584: Please add docstring

test/image-util-e2e-specs.js
```diff
@@ -61,6 +61,7 @@ describe('image-util', function () {
     let originalImage = null;
     let changedImage = null;
     let rotatedImage = null;
+    let numberImage = null;
```
mykola-mokhnach, line 64: imageIndex

test/image-util-e2e-specs.js
```diff
@@ -138,6 +140,32 @@ describe('image-util', function () {
         visualization.should.not.be.empty;
       });
     });
+
+    describe('getImageOccurrences', function () {
+      it('should return matches in the full image', async function () {
+        const occurrences = await getImageOccurrences(originalImage, numberImage, {threshold: 0.8});
+        occurrences.results.should.not.be.empty;
+        occurrences.results.length.should.be.eq(3);
```
mykola-mokhnach, line 148: the previous check is redundant

test/image-util-e2e-specs.js
```diff
@@ -138,6 +140,32 @@ describe('image-util', function () {
         visualization.should.not.be.empty;
       });
     });
+
+    describe('getImageOccurrences', function () {
```
mykola-mokhnach, line 144: 👍

lib/image-util.js
mykola-mokhnach: same below

lib/image-util.js
```diff
@@ -469,6 +469,124 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+      result.score = score;
+
+      result.rect = {
+        x: element.x,
+        y: element.y,
+        width: partialImg.cols,
+        height: partialImg.rows
+      };
+
+      results.push(result);
```
mykola-mokhnach, line 559:
```
results.push({
  score: ..,
  rect: {
    ...
  }
})
```

might be more readable

lib/image-util.js
KazuCocoa: Yes, it would be nice for debugging since the result could be different each other if lower threshold for example.

lib/image-util.js
khaong: Turns out getDataAsArray is quite slow; and all I needed to do is to use Mat.at() instead - so this disappears!

lib/image-util.js
khaong: It references the `minMax.maxVal` from inside the block; we could move it out but we'd need to declare a variable outside the try/catch scope

lib/image-util.js
khaong: I've changed it to return the visualisation per occurrence

KazuCocoa (approved):

lib/image-util.js
```diff
@@ -469,6 +470,120 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+    const matched = await fullImg.matchTemplateAsync(partialImg, cv.TM_CCOEFF_NORMED);
+    const minMax = await matched.minMaxLocAsync();
+
+    const nonZeroMatchResults = matched.threshold(threshold, 1, cv.THRESH_BINARY).convertTo(cv.CV_8U).findNonZero();
+    const matches = [];
+
+    for (const element of nonZeroMatchResults) {
+      let hasNoNeighbours = true;
+
+      matches.some((match) => {
```
mykola-mokhnach, line 528: `some` call returns `true` if any of the array items satisfies the given condition, so this whole block could be optimized

lib/image-util.js
```diff
@@ -469,6 +470,120 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+        const distanceBetweenPoints = distance(match, element);
+        if (distanceBetweenPoints <= MATCH_NEIGHBOUR_THRESHOLD) {
```
mykola-mokhnach, line 530: should we make MATCH_NEIGHBOUR_THRESHOLD value configurable?

mykola-mokhnach:
Could you please attach a visualized example with multiple matches?

lib/image-util.js
khaong: Is this what you mean?
```
    for (const element of nonZeroMatchResults) {
      if (!matches.some((match) => distance(match, element) <= MATCH_NEIGHBOUR_THRESHOLD)) {
        matches.push(element);
      }
    }
```

lib/image-util.js
khaong: 10 seems to get good results; I think the only time this will come into play is if you're searching for an image smaller than 10 pixels in size and the occurrences being very close together.

I'll put it in as an optional default.

lib/image-util.js
mykola-mokhnach: yes

lib/image-util.js
```diff
@@ -469,6 +470,120 @@ async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
...
+ * image.
+ *
+ * @param {Buffer} fullImgData The data of the full image packed into a NodeJS buffer
+ * @param {Buffer} partialImgData The data of the partial image packed into a NodeJS buffer
+ * @param {?OccurrenceOptions} options [{}] Set of occurrence calculation options
+ *
+ * @returns {OccurrenceResults}
+ * @throws {Error} If no occurences of the partial image can be found in the full image
+ */
+async function getImageOccurrences (fullImgData, partialImgData, options = {}) {
```
mykola-mokhnach, line 508: Perhaps we could combine this functionality with the existing occurrence finding helper by just adding another option, which allows to return array of occurrence results instead of a single occurrence. This way we'd need to do minimal changes to the higher level routines

khaong:
I've got a [local spec](https://gist.github.com/khaong/1cef8002dea2e49777a0664f3688b9dd) to generate the images

all:
full-highlights

individual:
highlight-0
highlight-1
highlight-2

lib/image-util.js
khaong: as well as the combined highlights

lib/image-util.js
khaong: I thought about that; and have a working base-driver implementation to make use of it. The higher level routines still need to handle both cases, so I'm not sure if putting it all in the single call is any easier?

I'm happy to change it if that's your guidance

mykola-mokhnach: a new function would require us to create a new API endpoint. Also, I can observe a lot of duplicateing code between the existing occurences lookup function and the new one. We anyway have an `options` argument, where we could add another flag to allow returning multiple occurrences. Also, if this flag is enabled we could just add another item to the returned object, which would contain a new item, which is the array of matched results. The visualisation remains the same

khaong: ok, have a look now

lib/image-util.js
```diff
@@ -435,38 +441,112 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
 async function getImageOccurrence (fullImgData, partialImgData, options = {}) {
   await initOpenCV();
 
-  const {visualize = false, threshold = DEFAULT_MATCH_THRESHOLD} = options;
+  const { visualize = false,
+          threshold = DEFAULT_MATCH_THRESHOLD,
+          multiple = false,
+          match_neighbour_threshold = MATCH_NEIGHBOUR_THRESHOLD } = options;
```
mykola-mokhnach, line 447: please change the param name to camelCase

lib/image-util.js
```diff
@@ -419,6 +423,8 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
  * region is highlighted with a rectangle.
  * @property {number} score The similarity score as a float number in range [0.0, 1.0].
  * 1.0 is the highest score (means both images are totally equal).
+ * @property {Array<OccurrenceResult>} multiple The array of matching OccurenceResults
```
mykola-mokhnach, line 426: The actual returned value is different and consists of `{results, visulalization}` items

lib/image-util.js
```diff
@@ -435,38 +441,112 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+      result.visualization = await cv.imencodeAsync('.png', singleHighlightedImage);
+    }
+    visualization = await cv.imencodeAsync('.png', fullHighlightedImage);
   }
-  return result;
+
+  return {
+    rect: results[0].rect,
+    score: results[0].score,
+    visualization: results[0].visualization,
```
mykola-mokhnach, line 514: I would rather put the visualization for multiple matches here. Then the `multiple` item could be an array

lib/image-util.js
```diff
@@ -435,38 +441,112 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+
+/**
+ * Filter out match results which have a matched neighbour
+ *
+ * @param {Mat} nonZeroMatchResults matrix of image match results
+ * @param {int} match_neighbour_threshold The pixel distance within which we
+ * consider an element being a neighbour of an existing match
+ * @return {Array<Point>} the filtered array of matched points
+ */
+function filterNearMatches (nonZeroMatchResults, match_neighbour_threshold) {
```
mykola-mokhnach, line 527: Please change match_neighbour_threshold to camelCase

lib/image-util.js
```diff
@@ -435,38 +441,112 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+  const matches = [];
```
mykola-mokhnach, line 528: this could be simplified to

```
return nonZeroMatchResults.reduce((acc, match) => {
  if (!matches.some((match) => distance(match, element) <= match_neighbour_threshold)) {
    acc.push(element);
  }
  return acc;
}, []);
```

lib/image-util.js
```diff
@@ -435,38 +441,112 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+    score: results[0].score,
+    visualization: results[0].visualization,
+    multiple: { results, visualization }
+  };
+}
+
+/**
+ * Filter out match results which have a matched neighbour
+ *
+ * @param {Mat} nonZeroMatchResults matrix of image match results
```
mykola-mokhnach, line 522: I assume this is an array

lib/image-util.js
khaong: Whoops, sorry! Bad rubyist 😓

lib/image-util.js
khaong: This is now actually correct with the changes to visualization below

lib/image-util.js
khaong: Thanks! I was trying to work out how to use a reduce

lib/image-util.js
```diff
@@ -408,6 +409,9 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
  * as an image (useful for debugging purposes)
  * @property {?float} threshold [0.5] At what normalized threshold to reject
  * a match
+ * @property {?float} multiple [false] find multiple matches in the image
+ * @property {?int} match_neighbour_threshold [10] The pixel distance between matches we consider
```
mykola-mokhnach, line 413: Please update the property name

mykola-mokhnach: I assume the distance could also be a float, so `number` is OK as its type

lib/image-util.js
```diff
@@ -435,38 +441,109 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+
   if (visualize) {
-    highlightRegion(fullImg, result.rect);
-    result.visualization = await cv.imencodeAsync('.png', fullImg);
+    const fullHighlightedImage = fullImg.copy();
+
+    for (const result of results) {
+      const singleHighlightedImage = fullImg.copy();
+
+      highlightRegion(singleHighlightedImage, result.rect);
```
mykola-mokhnach, line 504: do we still need singleHighlightedImage ?

lib/image-util.js
```diff
@@ -435,38 +441,109 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
     cv.imdecodeAsync(fullImgData),
     cv.imdecodeAsync(partialImgData)
   ]);
-  const result = {};
+  const results = [];
+  let visualization = null;
+
   try {
     const matched = await fullImg.matchTemplateAsync(partialImg, cv.TM_CCOEFF_NORMED);
     const minMax = await matched.minMaxLocAsync();
```
mykola-mokhnach, line 458: looks like we only need to calculate minMax for the `else` block

lib/image-util.js
```diff
@@ -435,38 +441,109 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
...
+          rect: {
+            x: element.x,
+            y: element.y,
+            width: partialImg.cols,
+            height: partialImg.rows
+          }
+        });
+      }
+    } else {
+      results.push({
```
mykola-mokhnach, line 476: here we should compare the score value with threshold before adding the match to results array

mykola-mokhnach: and then the condition below could be changed to `if(_.isEmpty(results))`

lib/image-util.js
khaong: It provides the individual per-occurrence highlighted result, as per this comment https://github.com/appium/appium-support/pull/212#discussion_r552198305

mykola-mokhnach: I would say in case of multiple matches it makes sense to highlight all matches on a single visualization

mykola-mokhnach: It looks like I was confused by variable names. Yes, the current logic is correct

lib/image-util.js
khaong: Yes, you are correct. In practice would someone define that distance as a fraction of pixels? 🤷 

Changed it in any case :)

lib/image-util.js
khaong: 👍

lib/image-util.js
```diff
@@ -419,6 +423,8 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
  * region is highlighted with a rectangle.
```
mykola-mokhnach, line 423: We probably should mention that this is going to have all matches highlighted when `multiple` is enabled

mykola-mokhnach (approved):

mykola-mokhnach:
Thanks @khaong , this was a long journey. 

Please don't forget to update https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/image-comparison.md after the whole implementation is completed

mykola-mokhnach:
Published in 2.50.0

khaong:
Thank you @mykola-mokhnach for your guidance and your patience with me! I have learnt a lot through the process 😃

PR #216, opened 2021-03-13
feat: Add a possibility to perform non-multipart file uploads

mykola-mokhnach:
https://discuss.appium.io/t/need-support-uploading-of-screen-recording-to-opened-s3/33254


lib/net.js
```diff
@@ -31,42 +31,49 @@ async function uploadFileToHttp (localFileStream, parsedUri, uploadOptions = {})
     fileFieldName = 'file',
```
KazuCocoa, line 31: Should be only `fileFieldName`? (Remove `= 'file'`)

mykola-mokhnach: That's correct. The property contains the name of the format field, which is supposed to carry the binary content of a file to be sent

KazuCocoa (approved):

PR #217, opened 2021-03-22
fix: Make sure we don't get an outdated screenshot if there was an error in MJPEG consumer

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/15209


KazuCocoa (approved):

jlipps (commented):
good idea thanks

Issue #220, opened 2021-04-03
undefined matchTemplateAsync


KazuCocoa:
https://github.com/appium/appium-support/blob/49e4ae237177fc746f1822dadaeef180e449c2e6/lib/image-util.js#L369
`@driver.find_image_occurrence` as well.

I confirmed this error in 5.1.0.
https://github.com/justadudewhohacks/opencv4nodejs/blob/5.1.0/test/tests/imgproc/MatImgprocTests.js#L847-L868

So, the method probably has changed in 5.x? (maybe we tested this before with 4.x. I could not remember tho)

KazuCocoa:
So far, this does not affect find_by_image.
occurrence and similarity have the method.


Quickly git grepped the differences. Probably something changed?
Then, we should raise an error for similarity and occurrence if no matched opencv4nodejs version.
```
HEAD is now at a006429 4.17.0
kazu$ git grep 'matchTemplateAsync'
cc/core/MatImgproc.cc:  Nan::SetPrototypeMethod(ctor, "matchTemplateAsync", MatchTemplateAsync);
lib/typings/Mat.d.ts:  matchTemplateAsync(template: Mat, method: number, mask?: Mat): Promise<Mat>;
test/tests/core/Mat/imgprocTests.js:      asyncFuncShouldRequireArgs(() => getImg().matchTemplateAsync());
test/tests/core/Mat/imgprocTests.js:        img.matchTemplateAsync(templ, cv.TM_SQDIFF_NORMED, (err, res) => {
test/tests/core/Mat/imgprocTests.js:        img.matchTemplateAsync(templ, cv.TM_SQDIFF, mask, (err, res) => {
...
kazu$ git checkout 5.1.0
Updating files: 100% (592/592), done.
Previous HEAD position was 5ab4fcb 4.3.0
HEAD is now at 4d585cc 5.1.0
kazu$ git grep 'matchTemplateAsync'
cc/imgproc/MatImgproc.cc:  Nan::SetPrototypeMethod(ctor, "matchTemplateAsync", MatchTemplateAsync);
lib/typings/Mat.d.ts:  matchTemplateAsync(template: Mat, method: number, mask?: Mat): Promise<Mat>;
test/tests/imgproc/MatImgprocTests.js:      asyncFuncShouldRequireArgs(() => getImg().matchTemplateAsync());
[...]

PR #221, opened 2021-04-04
chore: try/catch in matchTemplateAsync for error message

KazuCocoa:
Closes https://github.com/appium/appium-support/issues/220
When `matchTemplateAsync` fails, `getImageOccurrence` raises an error message which is `Cannot find any occurrences of the partial image in the full image`, meanwhile `getImagesSimilarity` returns no message. As a client, it only gets 'internal server error' with `undefined` error message.

It seems `e.message` by opencv4nodejs is usually `undefined`.

In addition to it, server-side message is like the below.

```
[debug] Support Loading local package 'opencv4nodejs'
[debug] [W3C (f7b505ad)] Encountered internal error running command: undefined
HTTP] <-- POST /wd/hub/session/f7b505ad-2e91-417b-86ea-e78ea94d1609/appium/compare_images 500 142 ms - 527
[HTTP]
```

So, i'd like to add more messages about no matched images there.
(Rarely the `e.message` had opencv error message, but I observed the case a few times in 10 to 20 times try.)

### DIFF
--- a/lib/image-util.js
+++ b/lib/image-util.js
@@ -370,7 +370,7 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
   try {
     matched = await reference.matchTemplateAsync(template, cv.TM_CCOEFF_NORMED);
   } catch (e) {
-    throw new Error(`Both images did not match. Original error: ${e.message}`);
+    throw new Error(`The reference image did not match to the template one. Original error: ${e.message}`);
   }
   const minMax = await matched.minMaxLocAsync();
   const result = {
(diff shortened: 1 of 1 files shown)


lib/image-util.js
```diff
@@ -366,7 +366,12 @@ async function getImagesSimilarity (img1Data, img2Data, options = {}) {
     reference.convertToAsync(cv.CV_8UC3)
   ]);
 
-  const matched = await reference.matchTemplateAsync(template, cv.TM_CCOEFF_NORMED);
+  let matched;
+  try {
+    matched = await reference.matchTemplateAsync(template, cv.TM_CCOEFF_NORMED);
+  } catch (e) {
+    throw new Error(`Both images did not match. Original error: ${e.message}`);
```
mykola-mokhnach, line 373: The reference image did not match to the template one

mykola-mokhnach (approved):

PR #222, opened 2021-04-29
feat: Make it possible to customize the method name for OpenCV-based template matching

mykola-mokhnach:
Related to https://discuss.appium.io/t/appium-opencv-images-differing-only-in-color-accept-as-equal/33567

@khaong Could you please help to review the PR?

### DIFF
--- a/lib/image-util.js
+++ b/lib/image-util.js
@@ -59,6 +59,33 @@ const AVAILABLE_MATCHING_FUNCTIONS = [
   'BruteForceSL2',
 ];
 
+const MATCHING_METHODS = [
+  'TM_CCOEFF',
+  'TM_CCOEFF_NORMED',
+  'TM_CCORR',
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):
seems good.
Perhaps we should add a settings API item like `getMatchedImageResult` for `visual` option in _find image element/s_(?)

mykola-mokhnach:
> Perhaps we should add a settings API item like getMatchedImageResult for visual option in find image element/s(?)

yes, makes sense

khaong:
Hi,

This looks good to me, I unfortunately haven't done enough testing with opencv itself to know how those other modes would interact with our code. I'll have a deeper look over the next few weeks to see if there are any potential issues.

I suggest we go ahead with the PR merge in any case for now.

ps thanks for fixing all the spelling mistakes 👍

PR #226, opened 2021-07-13
feat(zip): extractAllTo(): default to system unzip if present

boneskull:
For performance reasons with larger files (or perhaps just archives containing thousands of files), system `unzip` provides much better perf (~3x) than the native implementation.

Instead of simply preferring it over the native one, given option `useSystemUnzip: true` to `extractAllTo()`, we will now:

1. look in the `PATH` for `unzip` (cache the location)
2. if it's there, use it instead. if not, fall back to JS
3. if it fails for any reason, fall back to JS

If `useSystemUnzip` is falsy, we will use the JS implementation; the user need not change any code.

This strategy provides robust, opt-in behavior for better performance where needed.

The ultimate aim is for `appium-base-driver`/`@appium/base-driver` to leverage this behavior.

**This will need  to be backported (forward-ported?) to `@appium/support`**

Ref: appium/appium-base-driver#503 (which will become a PR that leverages this change)

### DIFF
--- a/lib/zip.js
+++ b/lib/zip.js
@@ -12,6 +12,7 @@ import { toReadableSizeString, GiB } from './util';
 import Timer from './timing';
 import log from './logger';
 import getStream from 'get-stream';
+import { exec } from 'teen_process';
 
 const openZip = B.promisify(yauzl.open);
 const pipeline = B.promisify(stream.pipeline);
@@ -179,6 +180,8 @@ class ZipExtractor {
  * @property {?number} defaultFileMode [0o644] The default permissions for extracted files. It is only
  * applied when the extractor is unable to retrieve this value for a file from the archive
  * metadata.
+ * @property {?boolean} useSystemUnzip [false] If true, attempt to use system unzip; if this fails,
+ * fallback to the JS unzip implementation.
  */
 
 /**
@@ -194,11 +197,39 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
   }
 
   await fs.mkdir(destDir, {recursive: true});
+  const dir = await fs.realpath(destDir);
+  if (opts.useSystemUnzip) {
+    try {
+      await extractWithSystemUnzip(zipFilePath, dir);
+      return;
+    } catch (err) {
+      log.warn('unzip failed; falling back to JS: %s', err.stderr || err.message);
+    }
+  }
   const extractor = new ZipExtractor(zipFilePath, {
     ...opts,
-    dir: await fs.realpath(destDir),
+    dir,
   });
   await extractor.extract();
+}
+
+/**
+ * Executes system unzip (e.g., `/usr/bin/unzip`).  If available, it is
+ * significantly faster than the JS implementation.
+ * This function will reject on win32.
+ * @param {string} zipFilePath The full path to the source ZIP file
+ * @param {string} destDir The full path to the destination folder
+ */
+async function extractWithSystemUnzip (zipFilePath, destDir) {
+  const unzipExecutablePath = await assertSystemUnzip();
+  // -q means quiet (no stdout)
+  // -o means overwrite
+  // -d is the dest dir
+  await exec(unzipExecutablePath, [
+    '-q',
+    '-o', zipFilePath,
+    '-d', destDir
+  ]);
 }
 
 /**
@@ -479,6 +510,23 @@ async function toArchive (dstPath, src = {}, opts = {}) {
     archive.finalize();
   });
 }
+
+/**
+ * Finds a system `unzip` executable. Rejects if it is not found.
+ */
+const assertSystemUnzip = _.memoize(
+  /**
+   * @returns {B<string>} Path to unzip executable
+   */
+  async () => {
+    try {
+      const unzipPath = await fs.which('unzip');
+      log.debug('found system unzip at %s', unzipPath);
+      return unzipPath;
+    } catch {
+      throw new Error('Could not find system unzip');
+    }
+  });
 
 export { extractAllTo, readEntries, toInMemoryZip, _extractEntryTo,
   assertValidZip, toArchive };
--- a/test/zip-e2e-specs.js
+++ b/test/zip-e2e-specs.js
@@ -8,184 +8,223 @@ import { MockReadWriteStream } from './helpers';
 chai.use(chaiAsPromised);
 
 describe('#zip', function () {
-  let assetsPath;
-  let zippedFilePath;
-  let tmpRoot;
 
-  beforeEach(async function () {
-    assetsPath = await tempDir.openDir();
-    tmpRoot = await tempDir.openDir();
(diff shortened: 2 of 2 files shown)


boneskull:
It is worth mentioning that spawning `unzip` will be _slower_ for smaller files.  But hopefully it won't matter much.  If it _does_ turn out to matter, we can restrict the behavior to "archive size _n_ in bytes" or "_m_ files in the archive" but that's probably being a little too smart.

lib/zip.js
```diff
@@ -194,11 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
 
   await fs.mkdir(destDir, {recursive: true});
-  const extractor = new ZipExtractor(zipFilePath, {
-    ...opts,
-    dir: await fs.realpath(destDir),
-  });
-  await extractor.extract();
+  const dir = await fs.realpath(destDir);
+  let unzipDone = false;
+  if (opts.useSystemUnzip && process.platform !== 'win32') {
```
mykola-mokhnach, line 202: you could use os.isWindows helper

lib/zip.js
```diff
@@ -194,11 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
+    try {
+      await execSystemUnzip(zipFilePath, dir);
```
mykola-mokhnach, line 204: could we also apply same options to this call?

lib/zip.js
```diff
@@ -194,11 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
+      unzipDone = true;
+    } catch (err) {
+      log.warn('unzip failed; falling back to JS: %s', err);
```
mykola-mokhnach, line 207: consider using string templates: 
```
log.warn(`unzip failed; falling back to JS: ${err.message}`);
```

lib/zip.js
```diff
@@ -194,11 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
-    ...opts,
-    dir: await fs.realpath(destDir),
-  });
-  await extractor.extract();
+  const dir = await fs.realpath(destDir);
+  let unzipDone = false;
+  if (opts.useSystemUnzip && process.platform !== 'win32') {
+    try {
+      await execSystemUnzip(zipFilePath, dir);
+      unzipDone = true;
```
mykola-mokhnach, line 205: we could simply perform short-circuit return here. There is no need for an additional flag variable

lib/zip.js
```diff
@@ -480,6 +509,23 @@ async function toArchive (dstPath, src = {}, opts = {}) {
   });
 }
 
+/**
+ * Finds a system `unzip` executable. Rejects if it is not found.
+ */
+const assertSystemUnzip = _.memoize(
```
mykola-mokhnach, line 515: I don't see a need for memoizing that. The `which` operation is fast, because it simply loops through the PATH content

lib/zip.js
```diff
@@ -480,6 +509,23 @@ async function toArchive (dstPath, src = {}, opts = {}) {
...
+  /**
+   * @returns {B<string>} Path to unzip executable
+   */
+  async function locateSystemUnzip () {
+    try {
+      const unzipPath = await fs.which('unzip');
+      log.debug('found unzip at %s', unzipPath);
+      return unzipPath;
+    } catch {
+      throw new Error('Could not find system unzip');
```
mykola-mokhnach, line 525: do we need to wrap the original error?

test/zip-e2e-specs.js
```diff
@@ -8,184 +8,191 @@ import { MockReadWriteStream } from './helpers';
 chai.use(chaiAsPromised);
 
 describe('#zip', function () {
```
mykola-mokhnach, line 10: it would probably make sense to add a test where we extract an archive containing unicode chars in file names

lib/zip.js
boneskull: it is needed for the fallback.  it's either that or duplicate the function call

lib/zip.js
boneskull: It's actually more performant to use printf-style interpolation, because it is not eager--it will only be computed when logging is enabled at the chosen level.

depending on what you're logging this can be significant.  so I always use interpolation when logging as a general rule.

lib/zip.js
boneskull: oh, wait, you're right.  I'm not sure what I was thinking here

lib/zip.js
boneskull: idk, seems safer to me.  `PATH` could be...very large.  `PATH` could contain stuff on an nfs or samba share.  who knows

lib/zip.js
boneskull: I prefer it since it adds a little more context -- we're looking for the system unzip -- instead of "Error: not found: unzip" which doesn't tell us quite as much

test/zip-e2e-specs.js
boneskull: do we have anything handy?

boneskull: (I'm also unsure what assertions should be made?)

mykola-mokhnach: testcase:
- create a zip archive with files containing unicode chars in their names 
- unzip the archive
- verify names are unchanged for the unzipped files

lib/zip.js
mykola-mokhnach: do you have a link to some article where I could read more about such optimization?

lib/zip.js
boneskull: I don't know what this is in reference to

lib/zip.js
```diff
@@ -194,11 +197,38 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
+ * significantly faster than the JS implementation.
+ * This function will reject on win32.
+ * @param {string} zipFilePath The full path to the source ZIP file
+ * @param {string} destDir The full path to the destination folder
+ */
+async function execSystemUnzip (zipFilePath, destDir) {
+  const unzipExecutablePath = await assertSystemUnzip();
+  // -o means overwrite
+  // -d is the dest dir
+  const result = await exec(unzipExecutablePath, ['-o', zipFilePath, '-d', destDir]);
```
mykola-mokhnach, line 230: it makes sense to make it less talkative by adding -q or even -qq options. This would also make it faster for archives with thousands of files and save the memory (exec stores the output into an internal buffer)

lib/zip.js
```diff
@@ -194,11 +197,38 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
+  log.verbose('%s result: %o', unzipExecutablePath, result);
```
mykola-mokhnach, line 231: here result will be a dictionary with the whole list of extracted files. Do we really want to print it?

lib/zip.js
boneskull: OK; I had it this way, but envisioned that the debugging output could be helpful.  It makes more sense to listen for `data` events on the child process streams if we do want to go that direction.  I suppose we can try to do that if needed (but unsure if `teen-process` provides this).

Will revert

lib/zip.js
boneskull: I mean... maybe?  it depends if `unzip` is doing weird things (like with unicode filenames).    Anyway, the `log.verbose()` call is free unless you're using that loglevel, but the dictionary will live in memory as you mentioned

boneskull: at any rate, I am going to make it quiet for now

lib/zip.js
mykola-mokhnach: I mean `defaultDirMode` and `defaultFileMode`. Basically we just need to tune chmod after extraction is completed and these options are provided. Otherwise we could simply remove them. I've checked Appium source and could find any place where these are used.

lib/zip.js
```diff
@@ -194,13 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
...
 }
 
+/**
+ * Executes system unzip (e.g., `/usr/bin/unzip`).  If available, it is
+ * significantly faster than the JS implementation.
+ * This function will reject on win32.
+ * @param {string} zipFilePath The full path to the source ZIP file
+ * @param {string} destDir The full path to the destination folder
+ */
+async function execSystemUnzip (zipFilePath, destDir) {
```
mykola-mokhnach, line 223: how about extractWithSystemUnzip

lib/zip.js
```diff
@@ -194,13 +197,37 @@ async function extractAllTo (zipFilePath, destDir, opts = {}) {
   }
 
   await fs.mkdir(destDir, {recursive: true});
+  const dir = await fs.realpath(destDir);
+  if (opts.useSystemUnzip) {
+    try {
+      await execSystemUnzip(zipFilePath, dir);
+      return;
+    } catch (err) {
+      log.warn('unzip failed; falling back to JS: %s', err);
```
mykola-mokhnach, line 206: we could show more detailed error message here. I usually use `err.stderr || err.message`

lib/zip.js
```diff
@@ -480,6 +507,23 @@ async function toArchive (dstPath, src = {}, opts = {}) {
   });
 }
 
+/**
+ * Finds a system `unzip` executable. Rejects if it is not found.
+ */
+const assertSystemUnzip = _.memoize(
+  /**
+   * @returns {B<string>} Path to unzip executable
```
mykola-mokhnach, line 515: missing `throws`  description

boneskull: is `@throws` appropriate when the function is `async`?  it will not throw, of course--it'll reject

mykola-mokhnach: yes, it is appropriate. Basically, from this perspective it is just important to describe the method contract to possible clients.

lib/zip.js
boneskull: weird... okay.  if someone asks for it we can add it.

lib/zip.js
boneskull: I think I disagree.  If this was a public API (it's not) and I was a consumer, this would be misleading to me; I will never get a thrown exception out of this function, and it would be pointless to try to catch one outside of an `await`.  

`async` or `Promise`-returning code can only reject.  These are two different behaviors, despite the muddiness of using `async/await` with `try/catch`.  

There's no official support for describing the types of rejected Promises in JSDoc _or_ TypeScript.  So I get that this is not ideal, but I think it's _also_ fair to say that any `Promise` should be treated as if it can reject.  That means using a `catch` somewhere.  

Perhaps `assertSystemUnzip` is not a good name, since "assert" implies a thrown exception. Unsure what would be obviously better, though.

boneskull:
I'd like to get this merged ASAP.  I've added a unicode-filename test case, and I think we've spun on this long enough...

mykola-mokhnach:
I'm ok to merge the PR as soon as CI passes

boneskull:
@mykola-mokhnach saved you the bother 😄

PR #227, opened 2021-07-22
feat: Add a possibility to use PowerShell to unzip files in Windows

mykola-mokhnach:
Since standalone unzip does not exist in Windows we could use PowerShell features to replace it: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-7.1


KazuCocoa (approved):

KazuCocoa:
oh, probably we need this in https://github.com/appium/appium/commits/2.0/packages/support as well

PR #233, opened 2021-11-17
fix: Only set Content-Length header for non-multipart uploads

mykola-mokhnach:
Related to https://github.com/appium/java-client/pull/1591

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -47,6 +47,7 @@ async function uploadFileToHttp (localFileStream, parsedUri, uploadOptions = {})
   if (axiosAuth) {
     requestOpts.auth = axiosAuth;
   }
+  requestOpts.headers = Object.assign({}, _.isPlainObject(headers) ? headers : {});
   if (fileFieldName) {
     const form = new FormData();
     form.append(fileFieldName, localFileStream);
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):

KazuCocoa (approved):

lib/net.js
```diff
@@ -62,12 +64,11 @@ async function uploadFileToHttp (localFileStream, parsedUri, uploadOptions = {})
...
+    Object.assign(requestOpts.headers, form.getHeaders());
     requestOpts.data = form;
   } else {
-    if (_.isPlainObject(headers)) {
-      requestOpts.headers = headers;
+    if (streamLength) {
+      Object.assign(requestOpts.headers, {'Content-Length': streamLength});
     }
     requestOpts.data = localFileStream;
   }
```
jaden-k, line 74: I think the code would be executed only if `fileFieldName` is `false`. 
Is there this case?
Should the client set `fileFiledName = false` for non-multipart uploads?
I was just curious, thanks~

mykola-mokhnach: yes, The idea is that if a client wants to have non-multipart upload then `fileFieldName` should be set to any falsy value (empty string/false/null)

PR #237, opened 2022-04-10
Removing moment in favor for date-fns

dmitrybirin:
Hello 👋 
Rather big change for small fix, but hear me out 🙏 

We're working with appium currently, and got some problem with the appium@2 regarding high [vurnability](https://snyk.io/vuln/npm:moment@2.24.0) in the moment.
I tried to found the source of it and found this library.
moment js is huge library currently, so I've decided to add something more lightweight for logging the timestamp
(more for example in [that comparison](https://blog.bitsrc.io/date-fns-vs-momentjs-9833f7463751))
Also added the test that checking timestamp

**Great alternative** 
just update the moment js, but maybe lets merge the test, though


mykola-mokhnach:
This package is no longer supported. Consider applying your PR to https://github.com/appium/appium/tree/2.0/packages/support instead

dmitrybirin:
Thanks for the info! I will. Closing this PR then.

